PR #13, opened 2021-03-04
Add ah_collection module

Tompage1994:
### What does this PR do?
Adds ability to directly push a collection to AH as a binary. Its not the best (e.g. it always returns changed) but thats mostly because the API for AH is not the best - it always return 202 (if it worked) and the data it returns is useless.

Run against AH

N/A

This leaves the collection waiting for approval. I aim to create a module to do an approval for the collection shortly


plugins/modules/ah_collection.py
```diff
@@ -0,0 +1,58 @@
...
+
+__metaclass__ = type
+
+
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+
+DOCUMENTATION = """
+---
+module: ah_collection
```
sean-m-sullivan: Everything looks good, want to test this, but may not have time this week. 

My only reservation is calling this ah_collection. Think it should maybe be ah_collection_upload. As it may be possible to manipulate other parts of collections in the future? or remove certain aspects of sets, that we may want to reserve ah_collection itself for those functions? What are your thoughts.

Tompage1994: Completely agree. Have just changed the name!

plugins/modules/ah_collection_upload.py
```diff
@@ -0,0 +1,58 @@
...
+        - Collection artifact file path.
+      required: True
+      type: str
+
+extends_documentation_fragment: redhat_cop.ah_configuration.auth
+"""
+
+
+EXAMPLES = """
+- name: Create Tower Ping job template
```
sean-m-sullivan: Ok, went through this again, and tested it, It works great. Before we merge can we change this example name to something better. Like Upload collection artifact, or Upload awx collection, or push, just not tower ping :)

As a round 2, we should add the ability to use a distribution path, and incorporate this into the publish role.

sean-m-sullivan (approved):
LGTM updated that task name

PR #18, opened 2021-03-19
Add ah_build module

Tompage1994:
### What does this PR do?
Adds a module to run an ansible-galaxy build. It pulls from the ansible modules which are used to run the ansible-galaxy CLI. 

Try using 

```
ansible -m redhat_cop.ah_configuration.ah_build localhost -c local [-a "force=true path=/path/to/collection output_path=/path/to/out"]
```

instead of 

```
ansible-galaxy collection install [-f --output-path /path/to/out /path/to/collection]
```

N/A

Linting will fail until #19 is merged. Please merge that first

### DIFF
--- a/plugins/modules/ah_build.py
+++ b/plugins/modules/ah_build.py
@@ -1,0 +1,77 @@
+#!/usr/bin/python
+# coding: utf-8 -*-
+
+# (c) 2020, Tom Page <@Tompage1994>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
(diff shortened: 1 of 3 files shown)


Tompage1994:
With this we will really be able to streamline the `publish` role to just use these modules rather than `command`

Tompage1994:
Linting will fail until #19 is merged due to previous linting issues

plugins/modules/ah_build.py
```diff
@@ -0,0 +1,76 @@
...
+      required: False
+      type: str
+      default: "."
+
+"""
+
+
+EXAMPLES = """
+- name: Approve redhat_cop.ah_configuration:v1.0.0
+  ah_approval:
```
sean-m-sullivan: So two things here, lets change the approve to build, and remove the version, and change the example module to ah_build

sean-m-sullivan (approved):
LGTM

PR #28, opened 2021-05-27
Fix ah_repository module for when no requirements are specified

Tompage1994:
### What does this PR do?
Fix to the ah_repository module where not specifying a requirements file caused a module error.

CI would now catch such an error (if it were working...)

N/A

N/A


crenwick93 (approved):
looks good to me

PR #35, opened 2021-06-24
retry approvals if failed as there can be a slight delay

Tompage1994:
### What does this PR do?
retries approvals on the publish role if failed as there can be a slight delay. I've seen this come up a few times and running the role again works and have seen the collection waiting for approval so must be caused by the delay of the collection making it to be ready for approval.


sean-m-sullivan (approved):
LGTM

PR #38, opened 2021-07-19
Add ability to use requirements file for repo

### LINKED ISSUE #25: Allow requirements_file option for ah_repository module
I did the `ah_repository` module pretty quickly last week and did the bare minimum to get it to work. It would be pretty cool if you could give it either a `requirements` option (the requirements in bare yaml as currently exists) or a `requirements_file` option which is just a file location containing a requirements file for the repository. You'd need to parse the file as text and can then pass it to the API as currently happens.

Tompage1994:
### What does this PR do?
Allows the use of `requirement_file` option which allows using a file rather than specifying the requirements in-line

Added a task to the testing playbook (although its currently broken)

resolves #25 

N/A

### DIFF
--- a/playbooks/collection_requirement_file.yml
+++ b/playbooks/collection_requirement_file.yml
@@ -1,0 +1,5 @@
+---
+- name: redhat_cop.ah_configuration
+- name: redhat_cop.tower_configuration
+- name: ansible.windows
+...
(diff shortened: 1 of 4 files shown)


sean-m-sullivan (approved):
Nice Update, Like the ability to use both,

PR #41, opened 2021-08-06
Adding modules for user and group management 

herve4m:
### What does this PR do?
This PR adds three new modules for managing users, groups, and group permissions in private automation hub:

* The `ah_user` module creates, updates, and deletes user accounts.
* The `ah_group` module creates and deletes user groups.
* The `ah_group_perm` module manages group permissions.

I believe that these modules cover all the functionalities that the web UI provides for user and group management.

The modules use the API as provided by the [Ansible Automation Hub UI](https://github.com/ansible/ansible-hub-ui) project.
The API is available through the `/api/galaxy/_ui/v1/` path and documented at `/pulp/api/v3/docs/`.
The API is not available on public Ansible Galaxy.

The`testing_playbook_user.yml` test playbook is provided under `playbooks`. All tasks must pass.
The variables at the top of the play must be updated to point to a private automation hub installation.

The playbook has been successfully tested against private automation hub v4.3.2 (AAP 2.0)

No issue. No problem to solve.

The module uses the UI API. The endpoints might also be exposed through the Pulp API, although the authentication and token management seem to be processed by the UI API (the `ah_token` module might not be usable if using the Pulp API). More research needed.


Tompage1994 (commented):
In general this looks good. Thanks for the contribution.

Do you know if these modules will also work with current versions of AH? (i.e. from AAP 1.2)? I don't have time to fully test this out right now but should be able to get to it the start of next week.

herve4m:
I've added support for AAP 1.2 and run the test on private automation hub version 4.2.5 (latest version in AAP 1.2)

herve4m:
### New Modules

I have created two modules for private automation hub, in addition to the `ah_user`, `ah_group`, and `ah_group_perm` modules.

Those two modules only work with AAP 2.0 because earlier versions do not provide execution environment (EE) image management.

#### The `ah_ee_namespace` Module

The `ah_ee_namespace` module manages execution environment namespaces.
The module can create and delete namespaces, and also associate group permissions to namespaces.

In the following container image name, the namespace is `ansible-automation-platform-20-early-access` (`ee-minimal-rhel8` is the repository and `2.0.0-10` is the tag):

```
registry.redhat.io/ansible-automation-platform-20-early-access/ee-minimal-rhel8:2.0.0-10
```

The web UI shows the execution environment namespaces under the `Container Registry` menu.
Notice that the web UI does not allow the deletion of namespaces, but the `ah_ee_namespace` module does.
The web UI cannot create namespaces because they are created when an image is pushed, by using `podman push` for example.

#### The `ah_ee_repository` Module

The  `ah_ee_repository` module manages execution environment repositories (called _containers_ in Pulp).
I used "repository" for the name to align with Quay's terminology (`namespace/repository:tag`).
The module manages the description and the README file associated with each repository.
The module can also delete repositories (the web UI does not provide that feature).

The module cannot create repositories (nor the web UI): you create repositories when you push container images, with `podman push` for example.

### APIs

Private automation hub provides several APIs (Pulp plug-ins):

* The `/api/galaxy/_ui/v1/` API comes from the Ansible Automation Hub UI project (https://github.com/ansible/ansible-hub-ui) and is used to manage users, groups, and some aspects of the `Container Registry`  component, such as the README file associated with execution environment repositories.
* The `/pulp/api/v3/` API comes with Pulp and the Pulp Container Plug-in.
* The `/api/galaxy/v3/` API comes with Pulp and the Pulp Ansible Plug-in.
  That API manages collections and is also available on Ansible Galaxy (I think that the other two APIs are only provided for private automation hub).

To manage all aspects of the `Container Registry`  component, the `ah_ee_namespace`  and  `ah_ee_repository` modules use the first two APIs.

### Authentication

For authentication, you can create a token that you can use in automation controller or `ansible.cfg` (`token` parameter) to install collection from Ansible Galaxy or private automation hub.
The `ah_token` module can create a token for you.

**Warning**: It seems that in private automation hub you can only have one token. If you generate a new token, then it replaces the preceding one. That breaks your automation controller or `ansible.cfg` configuration because the old token is not valid anymore.

In private automation hub, you cannot create a dedicated token to manage users, groups, and the container repository component.
Only login/password authentication is supported.
As a consequence, the `ah_user`, `ah_group`, `ah_group_perm`, `ah_ee_namespace`, and  `ah_ee_repository` modules cannot be used with a token (a user name and password must be provided for each call)

Tompage1994 (changes requested):
Sorry to take so long to get to this. 

It all looks good to me and seems to work. I've tested it on AAP2.0.

The only additional thing is that the testing playbook for ee_repository will break as there is the step to push a container to the repository.

Would it be possible to add a step at the start of the playbook to pull a container (probably from somewhere that doesnt need auth for ease) and push it to the registry. That way the CI testing will work (when we get it fixed!)

I would probably also like to change the ee_repository and ee_image so they make use of podman to pull in the images - but that doesn't need to be sorted now and this can be published as is.

Thanks

Tompage1994 (approved):
Thanks for that change. I suspect the CI will still fail because it may not have podman, but our CI is broken at the moment anyway so we can fix it when it comes to it.

Thanks for the contribution!

PR #58, opened 2022-02-13
added new group permisions

sean-m-sullivan:
### What does this PR do?
New Group permissions were added in the latest version of Automation Hub, added them as options
Also added the ability to delete namespaces using state: absent
Made change to existing item for delete to check that existing item exists, should mimic what works in[ awx.awx](https://github.com/ansible/awx/blob/60b6faff1963fb03cfdaa95c73a367fdae9f60b6/awx_collection/plugins/module_utils/controller_api.py#L637)


Tompage1994 (approved):
LGTM

PR #61, opened 2022-02-15
Add collection manage plugin

sean-m-sullivan:
### What does this PR do?
Add a collection manage plugin
        - If present will return data on a collection.
        - If present with version, will return data on a collection version.
        - If absent without version, will delete the collection and all versions.
        - If absent with version, will delete only specified version.

Automated tests

I wonder if I should merge in the functionality of ah_collection_upload to this, Upload if specified path, then get its data, and delete if state is absent.


plugins/modules/ah_collection_manage.py
```diff
@@ -0,0 +1,115 @@
...
+
+__metaclass__ = type
+
+
+ANSIBLE_METADATA = {"metadata_version": "1.1", "status": ["preview"], "supported_by": "community"}
+
+
+DOCUMENTATION = """
+---
+module: ah_collection_manage
```
Tompage1994: Not sure `_manage` fits in with usual naming for a module. This should potentially just be `ah_collection`

Tompage1994 (changes requested):
I like this. I haven't had a chance to test but will do before merging. 

I do think we should add the functionality of collection_upload into this and simply have one `ah_collection` module which does both things. I don't think it entirely makes sense to have `state: present` not actually cause an action (other than return data) since that doesn't fit in with creating a desired state. I would essentially just take the `ah_collection_upload` module and add the ability to delete collections. In other words `path` will probably be a required option when state == present

plugins/modules/ah_collection_manage.py
sean-m-sullivan: With the addition of the upload yes, I was trying to make it so it was differentiated.

sean-m-sullivan: Also thinking its going to be good to return the data anyways once its created, for reference, The data part, was more to have Something returned, why not get the collection data.

sean-m-sullivan:
updated to make the changes, integrate publish, and make it so it does best of both

plugins/modules/ah_collection.py
```diff
@@ -0,0 +1,135 @@
...
+      default: "present"
+      type: str
+
+extends_documentation_fragment: redhat_cop.ah_configuration.auth
+"""
+
+
+EXAMPLES = """
+- name: Upload collection to automation hub
+  ah_collection_upload:
```
Tompage1994: Should we not just remove the collection upload module completely?

Tompage1994 (changes requested):

Tompage1994 (changes requested):
This flow doesn't quite work if you specify the version and the path as the collection is awaiting approval and you get a failure of the task.

I think you should also do the check on version first to see if item exists, and if so then don't bother with attempting upload.

Tompage1994:
> This flow doesn't quite work if you specify the version and the path as the collection is awaiting approval and you get a failure of the task.
> 
> I think you should also do the check on version first to see if item exists, and if so then don't bother with attempting upload.

`ansible -m  redhat_cop.ah_configuration.ah_collection -a "path=redhat_cop-ah_configuration-0.0.1-test.tar.gz namespace=redhat_cop name=ah_configuration version='0.0.1-test'" localhost -c local` fails with:

```
localhost | FAILED! => {
    "changed": false,
    "msg": "Could not find Collection redhat_cop.ah_configuration with_version 0.0.1-test"
}
```

despite the collection having uploaded and awaiting approval

Tompage1994:
I wonder also if we should add the ability to auto-approve the collection in here too then implement the functionality of the ah_approval module here too.

Something like
```
if auto_approve:
  module.approve(
        endpoint=collection_endpoint,
    )
```

sean-m-sullivan:
@Tompage1994 Reworked for workflow, also added an override, so it can just force a new version if the user wants by using the delete, and reworked the entire flow, to incorporate the approval and override.

plugins/module_utils/ah_module.py
```diff
@@ -630,7 +630,8 @@ def wait_for_complete(self, task_url):
         if state == "failed":
             self.fail_json(msg="Upload of collection failed: {0}".format(response["json"]["error"]["description"]))
         else:
-            self.exit_json(**self.json_output)
+            time.sleep(1)
```
Tompage1994, line 633: I was going to complain about this, and perhaps we could do something such as provide an option for 'interval' and 'timeout' but I see this is exaclty what our friends in awx.awx do so I guess this is fine.

Tompage1994 (commented):
Will give a proper review later when I get time

plugins/module_utils/ah_module.py
sean-m-sullivan: I think I put in the sleep as between when it would upload, and the api would be populated it needed a second. Its for after the collection upload, not for waiting on it to be uploaded :)

plugins/modules/ah_collection.py
```diff
@@ -0,0 +1,169 @@
...
+    name: awx
+    path: /var/tmp/collections/awx_awx-15.0.0.tar.gz
+
+
+- name: Remove collection
+  ah_collection:
+    namespace: test_collection
+    name: test
+    version: 4.1.2
+    state: absent
```
Tompage1994, line 88: This doesnt work. If you add in a version you get the following:

`fatal: [localhost]: FAILED! => {"changed": false, "msg": "Unable to approve at collections/redhat_cop/ah_configuration/versions/0.0.8-dev: Awaiting approval not found"}`

plugins/modules/ah_collection.py
```diff
@@ -0,0 +1,169 @@
...
+"""
```
Tompage1994, line 89: Also 
```
- redhat_cop.ah_configuration.ah_collection:
        namespace: redhat_cop
        name: ah_configuration
        state: absent
        ah_host: https://localhost:8443
        ah_username: admin
        ah_password: password
        validate_certs: false
```

gives

`fatal: [localhost]: FAILED! => {"changed": false, "msg": "Could not find Collection redhat_cop.ah_configuration"}` when the collection isnt there

plugins/modules/ah_collection.py
```diff
@@ -0,0 +1,169 @@
...
+      description:
+        - Waits for the collection to be uploaded
+      type: bool
+      default: true
+    auto_approve
+      description:
+        - Approves a collection.
+        - Requires version to be set.
+      type: bool
+      default: true
```
Tompage1994, line 52: It seems the behaviour I've seen is that it defaults to false?

Tompage1994: Actually, seeing that you need to set version too. That being said the version doesn't seem to really be used. We could work out the version based on whats passed in (the old module did that)

plugins/modules/ah_collection.py
Tompage1994: i.e. not idempotent

Tompage1994 (changes requested):
Quite a few things and we are a fair way off accepting this

plugins/modules/ah_collection.py
sean-m-sullivan: Fixed for multiple scenarios with absent

plugins/modules/ah_collection.py
sean-m-sullivan: Changed behavior to just approve after upload if its set to auto approve.

plugins/modules/ah_collection.py
sean-m-sullivan: Fixed to be idempotent

sean-m-sullivan:
Sorry, busy week, Updated things, I think I've addressed the issues you found. I don't have an approval based AH atm, if you could test that? thanks

plugins/modules/ah_collection.py
```diff
@@ -0,0 +1,178 @@
...
+
+from ..module_utils.ah_module import AHModule
+
+
+def main():
+    # Any additional arguments that are not fields of the item can be added here
+    argument_spec = dict(
+        namespace=dict(required=True),
+        name=dict(required=True),
+        path=dict(),
```
Tompage1994, line 99: Could we have a properly thrown error when the path is not found. I currently get a nasty stack trace if providing a path which is incorrect

plugins/modules/ah_collection.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+    # Attempt to look up an existing item based on the provided data
+    if version:
+        collection_endpoint = "collections/{0}/{1}/versions/{2}".format(namespace, name, version)
+    else:
+        collection_endpoint = "collections/{0}/{1}".format(namespace, name)
+
+    existing_item = module.get_endpoint(collection_endpoint, **{"return_none_on_404": True})
+    if path:
```
Tompage1994, line 145: Also 

```
- redhat_cop.ah_configuration.ah_collection:
        namespace: redhat_cop
        name: ah_configuration
        version: 0.0.9-dev
        path: ../redhat_cop-ah_configuration-0.0.9-dev.tar.gz
        state: absent
        ah_host: https://localhost:8443
        ah_username: admin
        ah_password: password
        validate_certs: false
```
will upload the collection despite `state: absent`. Need to switch around this if statement here because if state is absent then things certainly shouldnt be created

Tompage1994 (changes requested):
Looks much better. One error-handling and one functional issue then all good.

plugins/modules/ah_collection.py
sean-m-sullivan: Moved state absent to be checked first and to exit module after its tasks.

plugins/modules/ah_collection.py
sean-m-sullivan: Add a pathlib check as that is supposed to be the module to use in python 3.4, on state = present to check for a path.

PR #63, opened 2022-03-02
Add ability to use remote path

### LINKED ISSUE #57: Add support for disconnected environments
This collection works really well to sync content from AH to PAH. Most of my customers operate in fully disconnected environments. Using the same collection to both pull content down from AH and then push that content to the customer's PAH in their environment would make disconnected operations much easier.

Tompage1994:
### What does this PR do?
Adds ability to use a remote path for a collection

```
- redhat_cop.ah_configuration.ah_collection_upload:
        path: https://galaxy.ansible.com/download/theforeman-foreman-3.0.0.tar.gz
```

resolves #57

@sean-m-sullivan Your PR #61 will want to pull in this change (it shouldn't conflict I think) and then you'll just need to add a line in the option of `path` that it accepts URLs

### DIFF
--- a/plugins/module_utils/ah_module.py
+++ b/plugins/module_utils/ah_module.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 from ansible.module_utils.basic import AnsibleModule, env_fallback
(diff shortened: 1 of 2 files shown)


sean-m-sullivan (approved):
Looks good!

PR #64, opened 2022-03-15
Sean m sullivan devel

sean-m-sullivan:
### What does this PR do?
Add a collection manage plugin
        - If present will return data on a collection.
        - If present with version, will return data on a collection version.
        - If absent without version, will delete the collection and all versions.
        - If absent with version, will delete only specified version.

Automated tests

I wonder if I should merge in the functionality of ah_collection_upload to this, Upload if specified path, then get its data, and delete if state is absent.


plugins/modules/ah_collection.py
```diff
@@ -0,0 +1,185 @@
...
+            existing_item = module.get_endpoint(collection_endpoint, **{"return_none_on_404": True})
+            if auto_approve:
+              module.approve(
+                    endpoint=collection_endpoint,
+                )
+        elif existing_item is None:
+            module.upload(path, "artifacts/collections", wait, item_type="collections")
+            module.json_output["changed"] = True
+            if auto_approve:
+              module.approve(
```
Tompage1994: I've run across this before and I'm running into it now. The collection is uploaded but the awaiting approval isn't ready when it gets to the code for approval. I think it may be worth adding a loop around the module.approve here to wait up to 5s or so for the approval to appear. (i.e. attempt approval and if fail sleep 1s). 

```
TASK [redhat_cop.ah_configuration.ah_collection] **********************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Unable to approve at collections/redhat_cop/ah_configuration/versions/0.0.9-dev: Awaiting approval not found"}
```
I can see following this that the collection is awaiting approval

Tompage1994 (changes requested):
One small change

Tompage1994:
Also, does this supersede the other PR open?

plugins/modules/ah_collection.py
sean-m-sullivan: Ok, will work on this, and yes this does supersede the other PR, #61 Just linking it in here now for reference and closing it.

Tompage1994 (approved):

PR #68, opened 2022-03-28
Ah hostname fix

### LINKED ISSUE #65: Confusing vars `ah_host` vs `ah_hostname`.
For using the collection I've defined `ah_hostname` as defined in the defaults, for instance (`roles/user/defaults/main.yml`) [states](https://github.com/redhat-cop/ah_configuration/blob/devel/roles/user/defaults/main.yml#L5):

`# ah_hostname: "{{ inventory_hostname }}"`

So I've defined my playbook as
```
- name: Playbook to configure automation hub after installing it
  hosts: all

  vars:
    ah_validate_certs: false
    ah_hostname: ah00
    galaxy_server: # omitted 
    ah_oauthtoken: !vault |
      # omitted
  collections:
    - redhat_cop.ah_configuration
  roles:
  [ omitted ]
```

Which fails complaining about the `ah_host` not being defined.
The README [does mention](https://github.com/redhat-cop/ah_configuration/blob/devel/roles/user/README.md?plain=1#L8) `ah_host` is needed, so maybe its just the `defaults/main.yml` that needs an update.

That said, in the [controller_configuration](https://github.com/redhat-cop/controller_configuration) the var is called `controller_hostname`. So for the sake of consistency perhaps is a better idea to change `ah_host` to `ah_hostname`.

Tompage1994:
### What does this PR do?
Changes `ah_hostname` to be the default option in roles to match controller_configuration. `ah_host` remains as an alias. Also added `ah_hostname` to be an alias for `ah_host` in modules

CI updated and an test running modules with either ah_host or ah_hostname and same with roles.

resolves #65 

N/A


sean-m-sullivan:
See I thought we'd just keep it the same as the module option accross the board, I'm not sure where we started tower_hostname, or how that came to be, but keeping it consistent between module/role was probably the best option.

Tompage1994:
ok, I've flipped it around and made the variable back to `ah_host` but made an alias to `ah_hostname` too which is the same for both roles and modules now.

sean-m-sullivan (approved):
Looks good,

PR #72, opened 2022-04-05
 Rename state in ee modules and create remote repositories 

### LINKED ISSUE #71: Default state of ah_ee_repository set to 'present' instead of 'updated'
In the `ah_ee_repository` module only `updated` and `absent` are allowed, with `updated` being the default. See [here](https://github.com/redhat-cop/ah_configuration/blob/devel/plugins/modules/ah_ee_repository.py#L172).
But in `roles/ee_repository/tasks/main.yml` state is defaulted to `present`, which causes the role to fail with a 
`"msg": "value of state must be one of: updated, absent, got: present"` if the state is not defined manually.

Tompage1994:
### What does this PR do?
Fixes the state option of the ah_ee_image and ah_ee_repository modules which were set as 'updated' or 'absent' when they should be 'present' or 'absent' as standard.

CI Updated

resolves #71 

N/A


Tompage1994:
Marking as draft as have found issues that these don't actually create the object. We should look into doing that

Tompage1994:
@sean-m-sullivan I've updated this.

I've done testing to the best of my limited brainpower and resources so this will require a bit more of a test to ensure it works but I think its good to go.

There is some weirdness around the repos (e.g. you cant rename a remote repository) although with some more trickery it might be possible (It doesn't seem possible for local repos in the UI but the pulp magic works in here). Can be something to add later on (perhaps raise as an issue for someone (or me) to work on at a later date?

Tompage1994:
@sean-m-sullivan I've made further changes so that you can now do `new_name` for remote repos :)

djdanielsson (approved):
LGTM

sean-m-sullivan (approved):
LGTM lets get it merged

PR #77, opened 2022-04-06
Create ah_ee_registry module

Tompage1994:
### What does this PR do?
Adds ability to create an EE Registry

### How should this be tested?
Added example to testing playbook, but also just running:

```
- name: Add EE Registry
      ah_ee_registry:
        name: myreg
        url: https://registry.redhat.io
```

### Is there a relevant Issue open for this?
resolves #76 

### Other Relevant info, PRs, etc.
N/A

PR #79, opened 2022-04-08
Add ah_ee_registry_sync module

### LINKED ISSUE #73: Create ah_ee_registry_sync

Tompage1994:
*Merge #77 first*

### What does this PR do?
Adds ah_ee_registry_sync module to the collection. It kicks off the task to sync all the container repos. Each registry has X container repos so this kicks off a task which has X child tasks. This tracks those and reports back when its all done (if wait set to true)

Not the easist thing to do but something like:

```
 - name: Ensure the registry description and README are set
      redhat_cop.ah_configuration.ah_ee_registry:
        name: rrio
        username: XXX
        password: YYY
        url: https://registry.redhat.io
        ah_hostname: https://localhost:8443
        ah_username: admin
        ah_password: password
        validate_certs: false

    - name: Sync the registry
      redhat_cop.ah_configuration.ah_ee_registry_sync:
        name: rrio
        interval: 5
        ah_hostname: https://localhost:8443
        ah_username: admin
        ah_password: password
        validate_certs: false
```
works (though it takes FOREVER for the syncs to happen - though that may be because my AH is very under-specced)

resolves #73 

Merge #77 first as this followed on because I needed some of the objects


plugins/modules/ah_ee_registry_sync.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/python
+# coding: utf-8 -*-
+
+# (c) 2020, Sean Sullivan <@sean-m-sullivan>
```
sean-m-sullivan: Fix the name

plugins/modules/ah_ee_registry.py
```diff
@@ -0,0 +1,176 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2021, Herve Quatremain <hquatrem@redhat.com>
```
sean-m-sullivan, line 4: name correct?

sean-m-sullivan (changes requested):
Need to do testing still, few minor changes needed.

sean-m-sullivan:
Looks and works great!, Lets fix the things I've noted and should be good to merge!

sean-m-sullivan (approved):
LGTM

PR #81, opened 2022-04-12
WIP: Remote container creation

sean-m-sullivan:
### What does this PR do?
Allow for the creation of containers from remote registries. 

This is my initial work on this, I am a bit lost as theres a pulp endpoint to get the href from, that is then used in an UI endpoint for creation. and you can't get the individual item from the UI. 

TBD

no


sean-m-sullivan:
@Tompage1994 if you want to take a look at this, its fun...

Tompage1994:
So I also had started working on this, but I was going to do the remote repositories under the same module as local repositories because in the UI it's the same thing. If you are going to split it put I think you can basically ignore the pulp part because it isn't really needed. 

Do you think from a functional perspective it's better to have `ah_ee_repository` and `ah_ee_remote_repository`?

sean-m-sullivan:
@Tompage1994 if you've started working on this, I'd say go on ahead, It make sense ounder ah_ee_repository with a switch, just was trying to figure this out after you did the remote stuff.

PR #89, opened 2022-04-27
Add async

Tompage1994:
### What does this PR do?
Adds Async job checking in the same way as controller_configuration

### How should this be tested?
`check it works`

### Is there a relevant Issue open for this?
Provide a link to any open issues that describe the problem you are solving.
resolves #88 

### Other Relevant info, PRs, etc.
N/A

PR #93, opened 2022-04-29
Multiple fixes for role issues

Tompage1994:
### What does this PR do?
After doing some more testing I found a load of issues (one of which was completely breaking everything and I'm not sure how I didnt catch it before... The key changes are:

- Fix issue where all roles had the wrong variable set and caused failures
- Fixed issue with new variables not included in ee_repository role
- Changed the default retries to 50 because I was occasionally finding the timeout being hit despite working. (Probably partially because of the fact that my AH is very underspecced)

Run a few of the roles

N/A

Broken since v0.7.0 based on my PR #89

### DIFF
--- a/changelogs/fragments/rolefixes.yml
+++ b/changelogs/fragments/rolefixes.yml
@@ -1,0 +1,5 @@
+---
+bugfixes:
+  - Fix issue where all roles had the wrong variable set and caused failures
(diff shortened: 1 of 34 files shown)


djdanielsson (approved):
I would guess this means we need to push 0.7.1 out today with these fixes

Tompage1994:
Yes, we will need to

PR #103, opened 2022-06-02
Fix broken else if

seansackowitz:
### What does this PR do?
else if statement is the same as the if, and so it would never be executed


sean-m-sullivan (approved):
This is a good catch, I think the else covered it, but good to fix this.

PR #104, opened 2022-06-06
Adding pah_configure usable template

tonyreveal:
### What does this PR do?
Adds a usable playbook, task files, and vars file for implementing ah_configuration.  Just populate the vars files and run the playbook with the correct tag(s).

Linting has already been done.  Otherwise, populate the vars file(s) and run the pah_configure.yml playbook with an appropriate tag (documented in README).

N/A.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -3,7 +3,7 @@ Private Automation Hub Configuration
 
 A single playbook and multiple task and vars files which can be used to define your Private Automation Hub configuration as code.  Update the vars files to define your objects and run the playbook to deploy your changes to your Private Automation Hub cluster(s).
 
-If executed with the `alltags` tag then the playbook will create all ojbects defined in all vars files in the appropriate order.
+If executed with the `alltags` tag then the playbook will create all objects defined in all vars files in the appropriate order.
 
 Use of some tags may require that you include other tags; for example if adding a collection but you haven't already added the correct collection namespace.
 
@@ -158,4 +158,4 @@ BSD
 Author Information
 ------------------
 
-An optional section for the role authors to include contact information, or a website (HTML is not allowed).
+[Tony Reveal](https://github.com/tonyreveal)
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,4 +1,4 @@
-Private Automation Hub Configuration
+Private Automation Hub Configuration Example Playbook
 =========
 
 A single playbook and multiple task and vars files which can be used to define your Private Automation Hub configuration as code.  Update the vars files to define your objects and run the playbook to deploy your changes to your Private Automation Hub cluster(s).
@@ -8,7 +8,6 @@ If executed with the `alltags` tag then the playbook will create all objects def
 Use of some tags may require that you include other tags; for example if adding a collection but you haven't already added the correct collection namespace.
 
 Available tags:
-- alltags
 - groups
 - users
 - namespaces
@@ -141,11 +140,15 @@ You can run this playbook from ansible cli or as a Job Template (future use case
 
 From the command line to define all objects:
 
-    ansible-playbook ah_configure.yml --tags alltags
+    ansible-playbook pah_configure.yml
 
     or just to create new job templates:
 
-    ansible-playbook ah_configure.yaml --tags namespaces
+    ansible-playbook pah_configure.yml --tags namespaces
+
+    or with ansible-navigator:
+
+    ansible-navigator run pah_configure.yml
 
 
 NOTE: If you do not specify any tags then no content will be created or updated.
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,161 @@
+Private Automation Hub Configuration
+=========
+
+A single playbook and multiple task and vars files which can be used to define your Private Automation Hub configuration as code.  Update the vars files to define your objects and run the playbook to deploy your changes to your Private Automation Hub cluster(s).
+
+If executed with the `alltags` tag then the playbook will create all ojbects defined in all vars files in the appropriate order.
+
+Use of some tags may require that you include other tags; for example if adding a collection but you haven't already added the correct collection namespace.
+
+Available tags:
+- alltags
+- groups
+- users
+- namespaces
+- publish
+- ee_images
+- ee_namespaces
+- registries
+- indices
+- regsync
+- repos
+- reposync
+
+Requirements
+------------
+
+This content utilizes the ah_configuration collection.  You will need connectivity to a Private Automation Hub server which has synchronized these collections or to the internet so that the collections can be installed.
+
+You will also need Private Automation Hub credentials with sufficient permissions to create the objects you define as code.  This will need to be a local account within the cluster and not an externally authenticated account.
+
+Variables
+--------------
+
+`pah_configure.yml`:
+
+    absent_present
+
+        I suggest leaving `absent_present` set to "present".
+
+`vars/ah_collection_namespaces.yml`:
+
+    ah_namespaces
+
+        Dictionary of collection namespaces to create.  One example exists that will need to be filled in.  You can also copy / paste the example for additional namespace types.
+
+`vars/ah_collection_publish.yml`:
+
+    ah_collections
+
+        Dictionary of collections to publish.  One example exists that will need to be filled in.  You can also copy / paste the example for additional collections.
+
+`vars/ah_ee_images.yml`:
+
+    ah_ee_images
+
+        Dictionary of execution environment images to create.  One example exists that will need to be filled in.  You can also copy / paste the example for additional execution environments.
+
+`vars/ah_ee_namespaces.yml`:
+
+    ah_ee_namespaces
+
+        Dictionary of EE namespaces to create.  One example exists that will need to be filled in.  You can also copy / paste the example for additional EE namespaces.
+
+`vars/ah_ee_registries.yml`:
+
+    ah_ee_registries
+
+        Dictionary of EE registries to create.  One example exists that will need to be filled in.  You can also copy / paste the example for additional EE registries.
+
+`vars/ah_ee_registry_indices.yml`:
+
+    ah_ee_registries
+
+        Dictionary of EE registries to index.  One example exists that will need to be filled in.  You can also copy / paste the example for additional EE registries to index.
+
+`vars/ah_ee_registry_sync.yml`:
+
+    ah_ee_registries
+
+        Dictionary of EE registries to synchronize.  One example exists that will need to be filled in.  You can also copy / paste the example for additional EE registries to sync.
+
+`vars/ah_ee_repositories.yml`:
+
+    ah_ee_repositories
+
+        Dictionary of EE Repositories to create.  One example exists that will need to be filled in.  You can also copy / paste the example for additional EE Repositories.
+
+`vars/ah_ee_repository_sync.yml`:
+
+    ah_ee_repositories
+
+        Dictionary of EE Repositories to sync.  One example exists that will need to be filled in.  You can also copy / paste the example for additional EE Repositories to sync.
+
+`vars/ah_groups.yml`:
+
+    ah_groups
+
+        Dictionary of Groups to create.  One example exists that will need to be filled in.  You can also copy / paste the example for additional Groups.
+
+`vars/ah_users.yml`:
+
+    ah_users
+
+        Dictionary of users to create.  One example exists that will need to be filled in.  You can also copy / paste the example for additional users.
+
+`vars/pah_vars.yml`:
+
+    ah_host:
+
+        Hostname of an Automation Hub cluster you wish to configure.
+
+    ah_username:
+
+        Username to access the Automation Hub defined as `ah_host`.
+
+    ah_password:
+
+        Password for the user defined as `ah_password`.
+
+    ah_token:
+
+        Rather than username and password you can provide a token to use instead.
+
+    validate_certs: false
+
+        Whether or not to validate the certificates when connecting to private automation hub.  This is false by default as Private Automation Hub installs with a self-signed certificate unless otherwise provided.
+
+    ah_path_prefix:
+
+        This is an optional variable.  Review the ah_configuration for information on it's use.  Omit if not needed.
+
+Dependencies
+------------
+
+You will need the `ah_configuration` collection.
+
+Playbook Execution
+----------------
+
+You can run this playbook from ansible cli or as a Job Template (future use case) in AAP's Controller.
+
+From the command line to define all objects:
+
+    ansible-playbook ah_configure.yml --tags alltags
+
+    or just to create new job templates:
+
+    ansible-playbook ah_configure.yaml --tags namespaces
+
+
+NOTE: If you do not specify any tags then no content will be created or updated.
+
+License
+-------
+
+BSD
+
+Author Information
+------------------
+
+An optional section for the role authors to include contact information, or a website (HTML is not allowed).
--- a/examples/README.md
+++ b/examples/README.md
@@ -3,8 +3,6 @@ Private Automation Hub Configuration Example Playbook
 
 A single playbook and multiple task and vars files which can be used to define your Private Automation Hub configuration as code.  Update the vars files to define your objects and run the playbook to deploy your changes to your Private Automation Hub cluster(s).
 
-If executed with the `alltags` tag then the playbook will create all objects defined in all vars files in the appropriate order.
-
 Use of some tags may require that you include other tags; for example if adding a collection but you haven't already added the correct collection namespace.
 
 Available tags:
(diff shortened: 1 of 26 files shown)


sean-m-sullivan (approved):
I like it, Its high time we had an example folder in Automation hub, I'm going to have Tom take a look over this before merging, but it Looks Awesome to me.

examples/pah_configure.yml
```diff
@@ -0,0 +1,95 @@
...
+
+  pre_tasks:
+    - name: Include Vars
+      include_vars: vars/pah_vars.yml
+
+  tasks:
+    - name: Include task to manage private automation hub groups
+      include_tasks: tasks/manage_groups.yml
+      tags:
+        - never
```
djdanielsson: I don't think I understand this, why are there tags called 'never'?

examples/pah_configure.yml
```diff
@@ -0,0 +1,95 @@
...
+  vars:
+    absent_present: "present"
+
+  pre_tasks:
+    - name: Include Vars
+      include_vars: vars/pah_vars.yml
+
+  tasks:
+    - name: Include task to manage private automation hub groups
+      include_tasks: tasks/manage_groups.yml
```
djdanielsson, line 20: why include a tasks file that imports a role? If I were to guess it has to do with tags, but this seems more complex then necessary for an example IMO.

djdanielsson (commented):
Also instead of including var files we could also just promote the idea of group_vars.

examples/pah_configure.yml
tonyreveal: This was a feature request by people who use my similar collection of playbook, tasks, and vars for controller_configuration.  They wanted the ability to run the playbook and select what objects to create rather that executing every collection role every time the playbook is executed.  This lets them only define a namespace or only define a user rather than running all collection roles just to add one user (for example).

examples/pah_configure.yml
tonyreveal: So that the main playbook remains in a clean condition and just executes the task file.  And any customization that needs to occur happens in the individual task file.  For example, maybe someone has a need to check namespaces for something, set some fact, and then run the namespace role from the ah_configuration collection.

It's also so that users of the same setup for controller_configuration are already familiar with the structure.

examples/pah_configure.yml
djdanielsson: interesting, while I can see a few people wanting this, I don't think it is for everyone. I am concerned about the complexity we are adding to an example in a way most people won't use.

examples/pah_configure.yml
djdanielsson: It would be an interesting poll to see how many people do something like this over using group_vars and a simple playbook that calls imports roles directly.

examples/pah_configure.yml
tonyreveal: What's so complex about it?  Just add your object definition to the correct vars file and run the playbook with the appropriate tag.

Think about it like this.  Do you want to wait 15 minutes for all of the other roles to finish running if all you wanted to do was a registry sync?  Or just execute the playbook with the regsync tag and get exactly the thing you wanted and nothing else?

examples/pah_configure.yml
tonyreveal: Elaborate.  You lost me at group_vars.  AFAIK group_vars is for ansible inventories and this particular task that you've selected is for creating user groups in Automation Hub.

And like I said - it's also formatted this way for familiarity / consistency with the stuff I've already contributed to the controller_configuration collection.

examples/README.md
```diff
@@ -0,0 +1,161 @@
+Private Automation Hub Configuration
+=========
+
+A single playbook and multiple task and vars files which can be used to define your Private Automation Hub configuration as code.  Update the vars files to define your objects and run the playbook to deploy your changes to your Private Automation Hub cluster(s).
+
+If executed with the `alltags` tag then the playbook will create all ojbects defined in all vars files in the appropriate order.
```
Tompage1994:
```suggestion
If executed with the `alltags` tag then the playbook will create all objects defined in all vars files in the appropriate order.
```

examples/README.md
```diff
@@ -0,0 +1,161 @@
...
+This content utilizes the ah_configuration collection.  You will need connectivity to a Private Automation Hub server which has synchronized these collections or to the internet so that the collections can be installed.
+
+You will also need Private Automation Hub credentials with sufficient permissions to create the objects you define as code.  This will need to be a local account within the cluster and not an externally authenticated account.
+
+Variables
+--------------
+
+`pah_configure.yml`:
+
+    absent_present
```
Tompage1994: Not sure I like this as a name. Maybe something like `target_state`?

examples/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+You can run this playbook from ansible cli or as a Job Template (future use case) in AAP's Controller.
+
+From the command line to define all objects:
+
+    ansible-playbook ah_configure.yml --tags alltags
+
+    or just to create new job templates:
+
+    ansible-playbook ah_configure.yaml --tags namespaces
```
Tompage1994: Is it worth also using the `ansible-navigator` example here too since that is where the community is headed

examples/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+License
+-------
+
+BSD
+
+Author Information
+------------------
+
+An optional section for the role authors to include contact information, or a website (HTML is not allowed).
```
Tompage1994: Can you add in your info here rather than leaving as the default text

examples/README.md
```diff
@@ -0,0 +1,161 @@
+Private Automation Hub Configuration
```
Tompage1994:
```suggestion
Private Automation Hub Configuration Example Playbook
```

examples/pah_configure.yml
Tompage1994: I'm not sure the tagging of never is really necessary here though? Surely if you just wanted to run the `manage_users` section you would just add the `users` tag and it would just include that and skip elsewhere. I'd expect not adding tags to run through everything rather than having to specify everything I do want it to run (unless I run with `--tags alltags` which is counterintuitive). I think removing that will end up with the same target state of being able to run a specific section but without the need of actually adding any tags if you don't want to.

examples/tasks/manage_collection_publish.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Include vars
+  include_vars: ../vars/ah_collection_publish.yml
+  tags:
+    - alltags
+    - publish
+
+- name: Import ah_configuration role
+  import_role:
+    name: redhat_cop.ah_configuration.
```
Tompage1994:
```suggestion
    name: redhat_cop.ah_configuration
```

examples/README.md
```diff
@@ -0,0 +1,161 @@
...
+You will need the `ah_configuration` collection.
+
+Playbook Execution
+----------------
+
+You can run this playbook from ansible cli or as a Job Template (future use case) in AAP's Controller.
+
+From the command line to define all objects:
+
+    ansible-playbook ah_configure.yml --tags alltags
```
Tompage1994:
```suggestion
    ansible-playbook pah_configure.yml --tags alltags
```

examples/README.md
Tompage1994:
```suggestion
    ansible-playbook pah_configure.yml --tags namespaces
```

examples/vars/pah_vars.yml
```diff
@@ -0,0 +1,8 @@
+---
+ah_host:
```
Tompage1994, line 2:
```suggestion
ah_hostname:
```

examples/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+`vars/ah_users.yml`:
+
+    ah_users
+
+        Dictionary of users to create.  One example exists that will need to be filled in.  You can also copy / paste the example for additional users.
+
+`vars/pah_vars.yml`:
+
+    ah_host:
```
Tompage1994, line 105:
```suggestion
    ah_hostname:
```

examples/pah_configure.yml
```diff
@@ -0,0 +1,95 @@
...
+  gather_facts: false
+  collections:
+    - redhat_cop.ah_configuration
+
+  vars:
+    absent_present: "present"
+
+  pre_tasks:
+    - name: Include Vars
+      include_vars: vars/pah_vars.yml
```
Tompage1994, line 14: This wont ever get included if including a tag.

```suggestion
      include_vars: vars/pah_vars.yml
      tags:
        - always
```

examples/vars/pah_vars.yml
```diff
@@ -0,0 +1,8 @@
...
+ah_username:
+ah_password:
+ah_token:
+validate_certs: false
```
Tompage1994:
```suggestion
ah_validate_certs: false
```

examples/vars/ah_collection_namespaces.yml
```diff
@@ -0,0 +1,16 @@
+# Refer to https://github.com/redhat-cop/ah_configuration/tree/devel/roles/namespace
+# for documentation and additional parameter information
+
+---
+ah_namespaces:
```
Tompage1994, line 5: I think I would prefer example values in all of these. Enough that I could run it without changing anything (except perhaps the `pah_vars` and get a decent demo environment set up

Tompage1994 (changes requested):

examples/vars/pah_vars.yml
tonyreveal: You sure?  The README for every role says `ah_host`.

examples/README.md
tonyreveal: You sure? The README for every role says ah_host.

examples/vars/pah_vars.yml
tonyreveal: You sure? The README for every role says `validate_certs`.

examples/README.md
Tompage1994: My bad. this should just be `ah_host`

examples/vars/pah_vars.yml
Tompage1994: Yeah, when I ran this it failed and I've just been through and checked and it should definitely be `ah_validate_certs`. I'll raise a doc PR to fix

examples/pah_configure.yml
djdanielsson: it's fine, it seems I am in the minority here. Once the changes Tom requested are done I will merge the PR

examples/pah_configure.yml
tonyreveal: Per [documentation](https://docs.ansible.com/ansible/latest/user_guide/playbooks_tags.html#selecting-or-skipping-tags-when-you-run-a-playbook), the default behavior is to run all tasks and ignore tags:

`--tags all - run all tasks, ignore tags (default behavior)`

I even tested this.  Hence the use of the never tag to ensure only the `always` and specific tag names execute when the playbook is run.

tonyreveal: Here I've removed the `never` and `alltags` tags and run the playbook without specifying a tag.  You can see it still tries to run each task.  Of course it fails because I have no data in the vars :

```
 ansible-playbook tonyreveal/pah_config_template/pah_configure.yml

PLAY [Playbook to configure Private Automation Hub] ***************************************************************************************************

TASK [Include Vars] ***********************************************************************************************************************************
Tuesday 07 June 2022  18:16:39 -0400 (0:00:00.023)       0:00:00.023 **********
ok: [localhost]

TASK [Include task to manage private automation hub groups] *******************************************************************************************
Tuesday 07 June 2022  18:16:39 -0400 (0:00:00.043)       0:00:00.067 **********
included: /home/tony/Ansible/tonyreveal/pah_config_template/tasks/manage_groups.yml for localhost

TASK [Include vars] ***********************************************************************************************************************************
Tuesday 07 June 2022  18:16:39 -0400 (0:00:00.062)       0:00:00.129 **********
ok: [localhost]

TASK [redhat_cop.ah_configuration.group : Add Automation Hub group] ***********************************************************************************
Tuesday 07 June 2022  18:16:39 -0400 (0:00:00.032)       0:00:00.162 **********
fatal: [localhost]: FAILED! => {"msg": "Invalid data passed to 'loop', it requires a list, got this instead: None. Hint: If you passed a list/dict of just one element, try adding wantlist=True to your lookup invocation or use q/query instead of lookup."}

PLAY RECAP ********************************************************************************************************************************************
localhost                  : ok=3    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```

tonyreveal: And then if I add the `never` tag back in and run the playbook with the `all` tag it will still skip the `never` tags, hence using the `alltags` tag:

```
ansible-playbook tonyreveal/pah_config_template/pah_configure.yml --tags all

PLAY [Playbook to configure Private Automation Hub] ***************************************************************************************************

TASK [Include Vars] ***********************************************************************************************************************************
Tuesday 07 June 2022  18:23:05 -0400 (0:00:00.028)       0:00:00.028 **********
ok: [localhost]

PLAY RECAP ********************************************************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

examples/README.md
tonyreveal: I don't see a change in this suggestion.

examples/README.md
tonyreveal: The var name really doesn't matter to me.  I just figured `absent_present` was pretty self explanatory as to what is expected for a value.  I can change it to `target_state` if preferred.

examples/README.md
tonyreveal: This is a good callout.  I'll update the README to include an ansible-navigator example as well.

examples/README.md
tonyreveal: Sure.

examples/pah_configure.yml
Tompage1994: Yes, thats my point. I think you should be able to run without tags and do everything. Otherwise running without tags does nothing. By removing `never` and `alltags` you can still specify one tag (e.g. `users`) and it'll do just that. I think thats slightly better functionality

examples/README.md
Tompage1994: Just a typo thats highlighted with `ojbects` vs `objects`

examples/pah_configure.yml
tonyreveal: Except that it was requested by a number of people to be written this way.  Do nothing unless a tag is passed.  I'm flexible.  I'm just explaining why it's written this way.

djdanielsson: I agree with Tom, if no tags are provided I think the expected result should be do everything.

tonyreveal: The change is made but I haven't pushed it.

tonyreveal:
@sean-m-sullivan or @Tompage1994 I believe I've made all of the requested changes.  It looks like the github action requires an approval to run.

Tompage1994:
I've kicked off the CI checks. I'll give it a review in the morning

examples/README.md
```diff
@@ -0,0 +1,164 @@
+Private Automation Hub Configuration Example Playbook
+=========
+
+A single playbook and multiple task and vars files which can be used to define your Private Automation Hub configuration as code.  Update the vars files to define your objects and run the playbook to deploy your changes to your Private Automation Hub cluster(s).
+
+If executed with the `alltags` tag then the playbook will create all objects defined in all vars files in the appropriate order.
```
Tompage1994: I think this line can be removed now as `alltags` is gone

```suggestion

```

examples/README.md
```diff
@@ -0,0 +1,164 @@
...
+    or just to create new job templates:
+
+    ansible-playbook pah_configure.yml --tags namespaces
+
+    or with ansible-navigator:
+
+    ansible-navigator run pah_configure.yml
+
+
+NOTE: If you do not specify any tags then no content will be created or updated.
```
Tompage1994: Again, this line can be removed now.

```suggestion

```

Tompage1994 (changes requested):
Just two small README changes.

When running its all good, though by having all the vars files with blank values it causes it to actually do things and fail in those modules. I wonder whether they should be set either as `[]` and have the options commented out or better yet have a full working demo example for each with bits that tie together? Happy to be left to the user though

examples/README.md
tonyreveal: Fixed.  Will push shortly

examples/README.md
tonyreveal: Fixed.  Will push shortly

djdanielsson (approved):
I think you fixed all the things that have been requested. I will let @Tompage1994 merge if he is good with everything now.

Tompage1994 (approved):

PR #106, opened 2022-06-07
Add error checking around 400 responses when a genuine error

Tompage1994:
### What does this PR do?
Add error checking around 400 responses when a genuine error comes ck from the API

I found an issue when trying to create a user and setting the password to `password`. The API complains that the password is too short and cant be a common word but the module just errors out because `response.read()` fails. Instead I've taken inspiration for what is done in `awx.awx` and just take that error and set it to response so that once the `response.read()` does fail it can display the error from the API

Fails out previously with 

```
- redhat_cop.ah_configuration.ah_user:
        username:         "myneweruser"
        password:         "password"
```
but now fails out but displays the error message.

Setting 
```
- redhat_cop.ah_configuration.ah_user:
        username:         "myneweruser"
        password:         "p@@ssword"
```
will work.

N/A

N/A

### DIFF
--- a/changelogs/fragments/response.yml
+++ b/changelogs/fragments/response.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - Fixed an issue where a genuine API error would cause a module to have an unhandled error.
+...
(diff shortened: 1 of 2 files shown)


djdanielsson (approved):
LGTM

sean-m-sullivan (approved):
LGTM

PR #107, opened 2022-06-07
fix readmes for validate_certs

Tompage1994:
### What does this PR do?
Fixes readmes which say `validate_certs` is the correct variable for roles but it should be `ah_validate_certs`

N/A - Doc change only

N/A

N/A


djdanielsson (approved):
LGTM

PR #108, opened 2022-06-16
adding key_path to auth with git as an option

djdanielsson:
### What does this PR do?
adding the ability to pull from a private repo with ssh key with a new key_path var and updated documentation. I also removed some white space.

This may want to be manually tested once to insure it works as expected.

none

none


Tompage1994 (approved):
LGTM

Issue #110, opened 2022-06-27
Fix the else in galaxy path prefix


sean-m-sullivan:
This code got broken, and it works for galaxy_ng and automation hub standard installs, but users on both can override the path prefix to be what ever they want such as mega_conglomerate_group_incorporated. 
The else right now will never trigger. 
https://github.com/redhat-cop/ah_configuration/blob/7eae301d2d265f6bed58e4166a9062bd3d07b0db/plugins/module_utils/ah_module.py#L127

In addition its been proposed we switch it so automation hub is the one that users will input if they want the official product, and galaxy if it is the upstream galaxy_ng. 

I don't like that proposal, as it conflicts with what the actual API is set to. 
However I think we should default galaxy path prefix variable to galaxy, to be in line with the product, so users of the official product don't have to even supply the path prefix, Only galaxy_ng and people overriding the path prefix will need to set that option.

In the end, It should be Switched, but at the product level, and we should match the product

Tompage1994:
I'm going to fix this logic up in one PR. We can open a second PR for doing wacky things with the path_prefix after if we choose...

djdanielsson:
so I like the plan for the default to work with automation hub. I still am kinda for the idea of swapping the names to more correctly point to what platform we are talking to not API endpoints because those are named wrong. I know this would be a code breaking change but it's so confusing that we say galaxy when we want automation hub and automation-hub for galaxy.

djdanielsson:
I do agree there is a good case to be said that if we default to automation hub (platform) then it should simplify users of our collection to look over other responses because they won't need to worry about setting anything there, it still just personally bothers me they are wrong at the platform level even if they are correct at the API naming level.

Tompage1994:
@sean-m-sullivan do we want to close this or do we want to have a further discussion of how to do the `path_prefix` stuff?

sean-m-sullivan:
With the defaulting to galaxy now, and the override for non product installs, I think we can close this,

PR #111, opened 2022-06-27
Update path_prefix setting

Tompage1994:
### What does this PR do?
Sorts out the path_prefix situation. The default is set at the top to galaxy so theres actually no point in any of the logic there.

Whilst I was at it I did a little refactoring of the auth which essentially copied the build_url logic.

Run any module and see it works 

related #110

### DIFF
--- a/plugins/doc_fragments/auth.py
+++ b/plugins/doc_fragments/auth.py
@@ -52,4 +52,5 @@ options:
     - For galaxy_ng this is either 'automation-hub' or the custom prefix used on install with GALAXY_API_PATH_PREFIX
     - For Automation Hub this is 'galaxy'
     type: str
+    default: 'galaxy'
 """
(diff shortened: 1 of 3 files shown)


sean-m-sullivan (approved):
Like the change, if you've tested this with any module, lets merge it. I'm assuming you have, but haven't tested it myself.

djdanielsson (commented):
should we not also set default to galaxy at least in this PR as well? oh wait, you are saying it is here https://github.com/redhat-cop/ah_configuration/blob/e03d10cc0d228784be56d7e900e5fefbf26f6be4/plugins/module_utils/ah_module.py#L51 ?

djdanielsson (approved):

Tompage1994:
> 

Correct it is set there. However I have just updated to add into the auth doc fragment that the default is `galaxy`

Tompage1994:
Screenshot 2022-06-28 at 09 11 50

PR #116, opened 2022-07-04
Fixed incorrect tasks names on serveral roles

### LINKED ISSUE #115: bad name?
so when testing some configs I am writing I noticed the output of this task says Create EE Image which I am pretty sure isn't correct.

https://github.com/redhat-cop/ah_configuration/blob/fda7ee4da64fd84e229b46b108781a49c2a11839/roles/group/tasks/main.yml#L25

Tompage1994:
### What does this PR do?
Fixes task names which were incorrectly "EE Image"

Just task name updates

resolves #115 

N/A

### DIFF
--- a/changelogs/fragments/tasknames.yml
+++ b/changelogs/fragments/tasknames.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - Fixed incorrect task names on serveral roles
+...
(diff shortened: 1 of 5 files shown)


djdanielsson (approved):
LGTM

PR #117, opened 2022-07-05
Repository sync role

inderpaltiwana:
### What does this PR do?
Creating a repository_sync role

Automated tests are preferred, but not always doable - especially for infrastructure. Include commands to run your new feature, and also post-run commands to validate that it worked. (please use code blocks to format code samples)

resolves #<number>

### DIFF
--- a/roles/repository_sync/tasks/main.yml
+++ b/roles/repository_sync/tasks/main.yml
@@ -5,7 +5,7 @@
   ah_repository_sync:
     name:                       "{{ __repository_item.name }}"
     wait:                        "{{ __repository_item.wait | default(omit)  }}"
-    auth_url:                   "{{ __repository_item.interval | default(1) }}"
+    interval:                   "{{ __repository_item.interval | default(1) }}"
     timeout:                      "{{ __repository_item.timeout | default(omit) }}"
   loop: "{{ ah_repositories }}"
   loop_control:
@@ -29,7 +29,7 @@
   loop_control:
     loop_var: __repositories_syncs_job_async_result_item
   when: __repositories_syncs_job_async_result_item.ansible_job_id is defined
-  no_log: "{{ ah_configuration_repository_syncs_secure_logging }}"
+  no_log: "{{ ah_configuration_repository_secure_logging }}"
   vars:
     ansible_async_dir: '/tmp/.ansible_async'
 ...
--- a/roles/repository_sync/tasks/main.yml
+++ b/roles/repository_sync/tasks/main.yml
@@ -2,7 +2,7 @@
 
 # Sync AH repository
 - name: Sync Automation Hub repository
-  ah_repository:
+  ah_repository_sync:
     name:                       "{{ __repository_item.name }}"
     wait:                        "{{ __repository_item.wait | default(omit)  }}"
     auth_url:                   "{{ __repository_item.interval | default(1) }}"
--- a/roles/repository_sync/tasks/main.yml
+++ b/roles/repository_sync/tasks/main.yml
@@ -1,0 +1,35 @@
+---
+
+# Sync AH repository
+- name: Sync Automation Hub repository
+  ah_repository:
+    name:                       "{{ __repository_item.name }}"
+    wait:                        "{{ __repository_item.wait | default(omit)  }}"
+    auth_url:                   "{{ __repository_item.interval | default(1) }}"
+    timeout:                      "{{ __repository_item.timeout | default(omit) }}"
+  loop: "{{ ah_repositories }}"
+  loop_control:
+    loop_var: "__repository_item"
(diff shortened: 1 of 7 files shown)


roles/repository_sync/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
+---
+
+# Sync AH repository
+- name: Sync Automation Hub repository
+  ah_repository_sync:
+    name:                       "{{ __repository_item.name }}"
+    wait:                        "{{ __repository_item.wait | default(omit)  }}"
+    auth_url:                   "{{ __repository_item.interval | default(1) }}"
```
djdanielsson: is this correct? auth_url doesn't seem correct here.

djdanielsson (changes requested):
I have a question about one line

roles/repository_sync/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+    timeout:                      "{{ __repository_item.timeout | default(omit) }}"
+  loop: "{{ ah_repositories }}"
+  loop_control:
+    loop_var: "__repository_item"
+  no_log: "{{ ah_configuration_repository_secure_logging }}"
```
djdanielsson, line 13: after a closer look, I am not sure we want this var for no_log I think you have it correct in the async job ah_configuration_repository_syncs_secure_logging

roles/repository_sync/tasks/main.yml
Tompage1994: Should be `interval`

roles/repository_sync/tasks/main.yml
Tompage1994: This is correct in the same way we have done for the other roles (e.g. `ee_repository_sync`). The idea here was that you would use the same var for the related roles.

Tompage1994 (changes requested):

roles/repository_sync/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+    jid: "{{ __repositories_syncs_job_async_result_item.ansible_job_id }}"
+  register: __repositories_syncs_job_async_result
+  until: __repositories_syncs_job_async_result.finished
+  retries: "{{ ah_configuration_repository_syncs_async_retries }}"
+  delay: "{{ ah_configuration_repository_syncs_async_delay }}"
+  loop: "{{ __repositories_syncs_job_async.results }}"
+  loop_control:
+    loop_var: __repositories_syncs_job_async_result_item
+  when: __repositories_syncs_job_async_result_item.ansible_job_id is defined
+  no_log: "{{ ah_configuration_repository_syncs_secure_logging }}"
```
Tompage1994: See what you mean. This one needs to change to `ah_configuration_repository_secure_logging`

```suggestion
  no_log: "{{ ah_configuration_repository_secure_logging }}"
```

roles/repository_sync/tasks/main.yml
Tompage1994: I see what you mean. Its the other line that's wrong. I've added a suggested line change

roles/repository_sync/tasks/main.yml
djdanielsson: No, ah_configuration_repository_secure_logging is already used in a role which is why I was trying to say I think it should be ah_configuration_repository_syncs_secure_logging instead

Tompage1994: Exactly, it should be the same var for the `repository` and `repository_sync` roles

Tompage1994: See that ee_registry_X uses the same idea:

https://github.com/redhat-cop/ah_configuration/blob/e1dfbab02ef95afd54f2d03c7c4bcb60988129f9/roles/ee_registry/tasks/main.yml#L26

https://github.com/redhat-cop/ah_configuration/blob/e1dfbab02ef95afd54f2d03c7c4bcb60988129f9/roles/ee_registry_index/tasks/main.yml#L18

roles/repository_sync/tasks/main.yml
inderpaltiwana: Yep didn't spot this before pushing up, will make relevant changes

roles/repository_sync/tasks/main.yml
inderpaltiwana: Updated with `ah_configuration_repository_secure_logging` as suggested

inderpaltiwana:
Have updated this PR based on the comments and suggestions from reviewers

Tompage1994 (approved):

Tompage1994:
LGTM. @djdanielsson Let me know if you disagree

djdanielsson (approved):
LGTM

PR #120, opened 2022-07-07
Add several options to ah_repository

Tompage1994:
### What does this PR do?
Added 6 new options to ah_repository module.

### How should this be tested?
I've tested against AAP2.1. Haven't tested 2.2 which is the `signed_only` option. But I can see in the upstream code that is the API option. If someone with a newer AH version than me could test that would be very helpful!

### Is there a relevant Issue open for this?
resolves #119 

### Other Relevant info, PRs, etc.
N/A

PR #124, opened 2022-07-07
rewrite of repository role

djdanielsson:
### What does this PR do?
After thinking about this for a while, because hub only allows for 2 repos I am not sure if the loop is the best way to go about it. So this is my suggested rewrite and open for feedback on the idea.

Automated tests should be fine

NA

#120 would need to be merged first


djdanielsson:
@Tompage1994 I am interested in hearing your thoughts on this

Tompage1994:
I very much get the premise behind this. I wonder whether there will ever be more repositories possible than rh-certified and community. I'm sure I remember some mention of abiltiies for promotion processes or similar to other AH instances on the roadmap which may happen through a repository endpoint. 

@sean-m-sullivan do you have any insight into this?

roles/repository/README.md
```diff
@@ -5,9 +5,8 @@ An Ansible Role to create Repositories in Automation Hub.
 ## Variables
```
Tompage1994, line 5: You'll need to mention

`ah_repository_certified` abd `ah_repository_community` here

roles/repository/tasks/main.yml
```diff
@@ -1,43 +1,50 @@
...
+- name: Update Community repository
+  ah_repository:
+    name:                 "community"
+    url:                  "{{ ah_repository_community['url'] | default('https://galaxy.ansible.com/api/') }}"
+    username:             "{{ ah_repository_community['username'] | default(omit) }}"
+    password:             "{{ ah_repository_community['password'] | default(omit) }}"
+    requirements:         "{{ ah_repository_community['requirements'] | default(omit) }}"
+    requirements_file:    "{{ ah_repository_community['requirements_file'] | default(omit }}"
+    download_concurrency: "{{ ah_repository_community['download_concurrency'] | default(10) }}"
+    rate_limit:           "{{ ah_repository_certified['rate_limit'] | default(8) }}"
```
Tompage1994: I assume these are all meant to be `ah_repository_community`

Tompage1994 (changes requested):

roles/repository/tasks/main.yml
djdanielsson: ooops, yep that is what i get for copy and pasting

roles/repository/README.md
djdanielsson: Ok I made a note about it, let me know if you think it should be changed or could be more clear

roles/repository/README.md
```diff
@@ -3,11 +3,11 @@
 An Ansible Role to create Repositories in Automation Hub.
 
 ## Variables
+These are the sub options for the vars ah_repository_certified and ah_repository_community which are dictionaries with the options you want. See examples for details.
```
Tompage1994:
```suggestion
These are the sub options for the vars `ah_repository_certified` and `ah_repository_community` which are dictionaries with the options you want. See examples for details.
```

roles/repository/README.md
Tompage1994: Looks good. Just added one suggestion to make it look a little nicer

Tompage1994 (changes requested):

Tompage1994 (commented):
In terms of the content here, I think this is good to go.

The question remains whether we want to accept this PR. I think for now this is good, but as I say, I have concerns we may need to backtrack in the future. @sean-m-sullivan I'll leave it in your hands

sean-m-sullivan:
@Tompage1994 That was my thought as well, we might have to add another one in the future, I don't think there will ever be something like Job templates where it is 20 of them, but in the end 3 or 4 at most ever. David Posted this to me earlier, I'd say if you approved of it Tom, that we could merge, I'm on the fence, but not against the change.

djdanielsson:
because each repo will have its own specific options that only work for that name i think it makes more since to split it up like this (even if a couple more get added in the future) then try and have the user understand it and correctly put it in a loop. it should also help with errors being a little more clear hopefully.

Tompage1994 (changes requested):
Ok, I'm happy.

Could you add a changelog to this though and make it set as a breaking_change

djdanielsson:
would it be a breaking change since we haven't actually released a version with this role yet?

djdanielsson:
also another thought I had, should I make the sync role mirror this new layout? and if so, what are your thoughts on making an option so it could call the sync from this role as well? @Tompage1994 @sean-m-sullivan

Tompage1994:
> would it be a breaking change since we haven't actually released a version with this role yet?

Its a breaking change to a previous version, so yes, I would say so. It's still worth being in the changelog as we should document properly what breaks when moving between versions

Tompage1994:
> also another thought I had, should I make the sync role mirror this new layout? and if so, what are your thoughts on making an option so it could call the sync from this role as well? @Tompage1994 @sean-m-sullivan

The sync role could definitely follow this track. I suppose it can't hurt to call the sync role from this role too if a `.sync` option is provided. I'd suggest all this to be a separate PR and separate issue

Tompage1994 (approved):

PR #128, opened 2022-07-08
fixing some repository role issues and updating repository sync role

djdanielsson:
### What does this PR do?
fixes some repository role issues and updating repository sync role to match new repository role

Manual test should be done once to insure everything works as expected.

none

#124


djdanielsson:
One thing that might need to be checked before merging if https://github.com/redhat-cop/ah_configuration/blob/05d38867acf94a46b9823059664d62d5f1e4dfb3/roles/repository/tasks/main.yml#L6 is correct. Also I wanted to note the reason why these are not async tasks because I found in my testing that it would fail, to make up for this I defaulted wait to false so the task wouldn't take long.

roles/repository/README.md
```diff
@@ -6,7 +6,7 @@ An Ansible Role to create Repositories in Automation Hub.
 These are the sub options for the vars `ah_repository_certified` and `ah_repository_community` which are dictionaries with the options you want. See examples for details.
 |Variable Name|Default Value|Required|Description|Example|
 |:---:|:---:|:---:|:---:|:---:|
-|`url`|"cloud.redhat or galaxy"|no|Remote URL for the repository.||
+|`url`|"cloud.redhat or galaxy"|no|Remote URL for the repository.|"https://galaxy.ansible.com/api/"|
```
Tompage1994: This isn't the default for ah_repository_certified. Maybe we need a separate column here for the default for each?

roles/repository/tasks/main.yml
```diff
@@ -31,7 +36,7 @@
     username:             "{{ ah_repository_community['username'] | default(omit) }}"
```
Tompage1994, line 36: Just realised we are missing a load of options here. E.g. auth_url and the cert stuff. I know by default we aren't authenticating against galaxy but there is always potential we might not be pointing at galaxy.ansible.com and instead at something else that may need auth. It's an option we can set in the UI so it should be an option we can set here (Obvioiusly set to `default(omit)`)

roles/repository_sync/tasks/main.yml
```diff
@@ -1,35 +1,65 @@
...
-  retries: "{{ ah_configuration_repository_syncs_async_retries }}"
-  delay: "{{ ah_configuration_repository_syncs_async_delay }}"
-  loop: "{{ __repositories_syncs_job_async.results }}"
-  loop_control:
-    loop_var: __repositories_syncs_job_async_result_item
-  when: __repositories_syncs_job_async_result_item.ansible_job_id is defined
-  no_log: "{{ ah_configuration_repository_secure_logging }}"
-  vars:
-    ansible_async_dir: '/tmp/.ansible_async'
+# - name: Sync Automation Hub repository
```
Tompage1994: Definitely shouldn't have this commented code

roles/repository_sync/tasks/main.yml
```diff
@@ -1,35 +1,65 @@
...
     loop_var: "__repository_item"
   no_log: "{{ ah_configuration_repository_secure_logging }}"
-  async: 1000
-  poll: 0
-  register: __repositories_syncs_job_async
-  changed_when: not __repositories_syncs_job_async.changed
   vars:
     ansible_async_dir: '/tmp/.ansible_async'
+    _ah_repositories: "{{ _certified + _community }}"
+    _community: "{{ lookup('template', 'community.j2') }}"
```
Tompage1994: This is nasty. I really don't like this. It would just be better to play out both tasks as their own bit rather than hacking around to get it one task with a loop. The `item=...` message would be horrendous on this I imagine.

roles/repository_sync/defaults/main.yml
```diff
@@ -7,16 +7,7 @@
...
-# possible fields:
-#  - name: "namespace_name" # mandatory
-#    interval: 1
-#    wait: true
-#    timeout:
-
 ah_configuration_repository_secure_logging: "{{ ah_configuration_secure_logging | default(false) }}"
-ah_configuration_repository_async_retries: "{{ ah_configuration_async_retries | default(50) }}"
-ah_configuration_repository_async_delay: "{{ ah_configuration_async_delay | default(1) }}"
+# ah_configuration_repository_syncs_async_retries: "{{ ah_configuration_async_retries | default(50) }}"
```
Tompage1994: Please don't just leave commented out stuff in here

Tompage1994 (changes requested):

Tompage1994:
> 

I guess it fails because it takes ages. We could presumably default the retries to a much higher number?

djdanielsson:
no it instantly fails for me

roles/repository_sync/defaults/main.yml
djdanielsson: i left the async stuff commented in if we figured out how to get it working

roles/repository_sync/tasks/main.yml
djdanielsson: the item looks normal it will return either certified or community i am pretty sure

roles/repository_sync/tasks/main.yml
djdanielsson: the async code wasn't working for me which is why it is all commented out

roles/repository/tasks/main.yml
djdanielsson: according to the UI there is no auth_url for community. I believe I have all the options that are in the UI for each type.

roles/repository/README.md
djdanielsson: that is why i said in default value it was either cloud.redhat or galaxy and the example was the full url to show that but i understand that isn't the most clear. I wasn't sure how to make it clear having 2 of the same lines though

roles/repository_sync/tasks/main.yml
djdanielsson:
```
TASK [redhat_cop.ah_configuration.repository_sync : Sync Automation Hub repository] **********************************************************************************************************************************************************
changed: [hub] => (item={'name': 'rh-certified'}) => {"__repository_item": {"name": "rh-certified"}, "ansible_loop_var": "__repository_item", "changed": true, "task": "bc75a278-6632-4e55-bae8-eadb7a70f0dd"}
changed: [hub] => (item={'name': 'community'}) => {"__repository_item": {"name": "community"}, "ansible_loop_var": "__repository_item", "changed": true, "task": "7ab8727a-830c-43f2-9aa4-d774e865853e"}
```
this is the output of the role with -v

Tompage1994: Oh, you're right.

Either way, I really don't like the fact you use a template to build out the options for a module. Couldn't you just do it by setting vars and then do a lookup using the `vars` lookup?

roles/repository/tasks/main.yml
Tompage1994: You're right. Not sure what happens if you set it with the api. That said you can definitely set client_key etc. in there so that should be added

roles/repository/README.md
Tompage1994: Yes, but the the user won't see exactly what the default really is. We should put the full values in there

roles/repository_sync/tasks/main.yml
djdanielsson: If you have a better way then please post an example, this was the only way I could think of that wouldn't get set as "" because omit doesn't work outside of modules.

Tompage1994: What about something like this:

```
---
- name: Sync Automation Hub repository
  ah_repository_sync:
    name:           "{{ __repository_item.name }}"
    wait:           "{{ __repository_item.wait | default(false)  }}"
    interval:       "{{ __repository_item.interval | default(1) }}"
    timeout:        "{{ __repository_item.timeout | default(omit) }}"
    ah_host:        "{{ ah_host | default(ah_hostname) }}"
    ah_username:    "{{ ah_username | default(omit) }}"
    ah_password:    "{{ ah_password | default(omit) }}"
    ah_path_prefix: "{{ ah_path_prefix | default(omit) }}"
    ah_verify_ssl:  "{{ ah_validate_certs | default(omit) }}"
  when: _ah_repositories != []
  loop: "{{ _ah_repositories }}"
  loop_control:
    loop_var: "__repository_item"
  no_log: "{{ ah_configuration_repository_secure_logging }}"
  vars:
    ansible_async_dir: '/tmp/.ansible_async'
    _ah_repositories:
      - "{{ ah_repository_certified }}"
      - "{{ ah_repository_community }}"
```

roles/repository/tasks/main.yml
djdanielsson: I think I have that... I just looked at the file in this commit and both tasks have client_key

roles/repository_sync/tasks/main.yml
djdanielsson: but I have to add .name to these lists so this wouldn't work

djdanielsson: I can try and see if i can get something like this to work but I don't know if it will
```
  vars:
    ansible_async_dir: '/tmp/.ansible_async'
    _ah_repositories:
      - "{{ ah_repository_certified + {'name': 'rh-certified'} }}"
      - "{{ ah_repository_community + {'name': 'community'} }}"
```

roles/repository/tasks/main.yml
Tompage1994: Ugh... ignore me. The diff didn't expand properly the last time I viewed the file

roles/repository_sync/tasks/main.yml
Tompage1994: Good point...

This should do the trick

```
_ah_repositories:
          - "{{ ah_repository_certified | combine({'name': 'rh_certified'}) }}"
          - "{{ ah_repository_community | combine({'name': 'community'}) }}"
```

Tompage1994: And you can deal with the loop label as such:

```
      loop: "{{ _ah_repositories }}"
      loop_control:
        loop_var: "__repository_item"
        label: "{{ __repository_item.name }}"
      vars:
        ansible_async_dir: '/tmp/.ansible_async'
        _ah_repositories:
          - "{{ ah_repository_certified | combine({'name': 'rh_certified'}) }}"
          - "{{ ah_repository_community | combine({'name': 'community'}) }}"
```

djdanielsson: I am still working on this, it looks like it should work fine but I am getting an error so trying to understand why.

roles/repository_sync/defaults/main.yml
djdanielsson: I can delete if you really want

roles/repository/README.md
djdanielsson: take a look and see if you think this looks good now

djdanielsson:
also should we update the default on no_log to true for this? because it will show the full cloud token in the output which should be considered confidential

roles/repository/tasks/main.yml
```diff
@@ -3,7 +3,7 @@
 - name: Update Certified repository
   ah_repository:
     name:                 "rh-certified"
-    url:                  "{{ ah_repository_certified['url'] | default('https://cloud.redhat.com/api/automation-hub/') }}"
```
Tompage1994: Why have you removed this default? Does this URL not work anymore?

roles/repository/README.md
```diff
@@ -1,23 +1,27 @@
...
-|`requirements_file`|""|no|A yaml requirements file to download from remote.||
+|`requirements`|""|no|(`ah_repository_community`)Requirements to download from remote.||
+|`requirements_file`|""|no|(`ah_repository_community`)A yaml requirements file to download from remote.||
 |`proxy_url`|""|no|Proxy URL to use for the connection.||
 |`proxy_username`|""|no|Proxy URL to use for the connection.||
 |`proxy_password`|""|no|Proxy URL to use for the connection.||
 |`download_concurrency`|"10"|no| Number of concurrent collections to download.||
-|`rate_limit`|"8"|Limits total download rate in requests per second||
+|`rate_limit`|"8"|no|Limits total download rate in requests per second||
 |`signed_only`|""|no|Only download signed collections|True|
```
Tompage1994: Some of the boolean values are defaulted to false here. Could you update. E.g. `signed_only` defaults to false.

Tompage1994 (changes requested):
Looking much better. Just tiny things now and one question.

roles/repository/tasks/main.yml
djdanielsson: That is my question, it didn't seem to work so that is why I removed it. If it does still work I will put it back in

Tompage1994: Just tried doing a sync with `https://cloud.redhat.com/api/automation-hub/` in AH and it does work, so I would say leave that as default, then if a user wants to use a synclist they can

roles/repository/README.md
djdanielsson: I think I got them all defaulted now

roles/repository/tasks/main.yml
djdanielsson: cool, added back in

Tompage1994 (approved):
I'm happy with this now I think 👍

Tompage1994:
Would be good if @sean-m-sullivan could give it a once over to make sure everything looks good here

sean-m-sullivan (approved):
LGTM

PR #131, opened 2022-07-19
adding ssh_opts

djdanielsson:
### What does this PR do?
adding ssh_opts so users can have more power when pulling git repos such as 'StrictHostKeyChecking=no'

I am not sure we really need to test this because it is a builtin module option

none

none


sean-m-sullivan (approved):
LGTM

PR #132, opened 2022-07-27
Add generic API lookup plugin

Tompage1994:
### What does this PR do?
Adds a generic API lookup plugin to the collection similar to https://docs.ansible.com/ansible/latest/collections/awx/awx/controller_api_lookup.html

The code was lifted from the awx.awx collection and modified as appropriate

I've tested against api, _ui and pulp type endpoints and have ensured working with local and keycloak users for each.

I used this playbook which gave a good overview of the params too.
```
- hosts: localhost
  gather_facts: false
  tasks:
    - debug:
        msg: "{{ lookup('redhat_cop.ah_configuration.ah_api', 'ee_namespaces', host='https://ah.local', username='admin', password='password', verify_ssl=false, return_all=true) | map(attribute='name') | list }}"
    - debug:
        msg: "{{ lookup('redhat_cop.ah_configuration.ah_api', 'users', host='https://ah.local', username='admin', password='password', verify_ssl=false, query_params={'username__startswith': 'a'}) }}"
    - debug:
        msg: "{{ lookup('redhat_cop.ah_configuration.ah_api', 'namespaces', host='https://ah.local', username='admin', password='password', verify_ssl=false, query_params={'company': 'Community Test'}) }}"
    - debug:
        msg: "Users: {{ lookup('redhat_cop.ah_configuration.ah_api', 'ee_namespaces', host='https://ah.local', username='ansible', password='password', verify_ssl=false, return_all=true) | map(attribute='name') | list }}"
    - debug:
        msg: "{{ lookup('redhat_cop.ah_configuration.ah_api', 'users', host='https://ah.local', username='ansible', password='password', verify_ssl=false, return_all=true) | map(attribute='username') | list }}"
    - debug:
        msg: "{{ lookup('redhat_cop.ah_configuration.ah_api', 'namespaces', host='https://ah.local', username='ansible', password='password', verify_ssl=false, return_all=true) | map(attribute='name') | list }}"
```

N/A

A customer pointed out there was no easy way to get info from the AH before doing additional stuff without going through the API and having to deal with the auth etc. This will solve that.

### DIFF
--- a/plugins/lookup/ah_api.py
+++ b/plugins/lookup/ah_api.py
@@ -17,17 +17,17 @@ options:
     description:
       - The endpoint to query.
     choices:
-      - ee_images (A second term of the image repository name is required)
-      - ee_namespaces
-      - ee_registries
-      - ee_repositories
-      - collections
-      - collection (Three additional terms are required with the repository {published, rh_certified, community}, the namespace, and the collection name)
-      - groups
-      - namespaces
-      - repository_community
-      - repository_rh_certified
-      - users
+      - 'ee_images', [ee_image_repository_name]
+      - 'ee_namespaces'
+      - 'ee_registries'
+      - 'ee_repositories'
+      - 'collections'
+      - 'collection', [repository={published, rh_certified, community}], [collection_namespace], [collection_name]
+      - 'groups'
+      - 'namespaces'
+      - 'repository_community'
+      - 'repository_rh_certified'
+      - 'users'
     required: True
   query_params:
     description:
@@ -74,6 +74,7 @@ options:
 extends_documentation_fragment: redhat_cop.ah_configuration.auth_plugin
 notes:
   - If the query is not filtered properly this can cause a performance impact.
+  - Two options take multiple terms. 'ee_images' and 'collection'. See the _terms choices above or the examples below for more details.
 """
 
 EXAMPLES = """
--- a/plugins/lookup/ah_api.py
+++ b/plugins/lookup/ah_api.py
@@ -17,11 +17,12 @@ options:
     description:
       - The endpoint to query.
     choices:
-      - ee_images
+      - ee_images (A second term of the image repository name is required)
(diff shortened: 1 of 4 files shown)


plugins/lookup/ah_api.py
```diff
@@ -0,0 +1,252 @@
...
+- name: Report the usernames of all users
+  debug:
+    msg: "Users: {{ query('redhat_cop.ah_configuration.ah_api', 'users', return_all=true) | map(attribute='username') | list }}"
+
+- name: List all collection namespaces by the devops team
+  debug:
+    msg: "{{ lookup('redhat_cop.ah_configuration.ah_api', 'namespaces', host='https://ah.example.com', username='ansible',
+              password='Passw0rd123', verify_ssl=false, query_params={'company': 'Devops'}) }}"
+
+- name: Get the list of tags for my_ee
```
sean-m-sullivan, line 90: I think there needs to be some more documentation around these two and the options that are required, I know you put in notes above, but think its slightly confusing.

sean-m-sullivan (changes requested):
LGTM I just think needs some additional documentation, This is awesome and powerful!

plugins/lookup/ah_api.py
Tompage1994: I can add it to the notes section, but I'm not entirely sure how better to document it other than to add to the notes to look at the examples.

Tompage1994: Pushed an extra note and perhaps made the choices more (or less?) clear

sean-m-sullivan: It's been clear to me, Just worried that once people discover this feature they are going to create issues 'not' understanding it, this is good as is, we can figure out more documentation later, maybe a readme spelling it out for people not used to lookup plugins :)

sean-m-sullivan (approved):

PR #141, opened 2022-08-12
Testing Update

sean-m-sullivan:
### What does this PR do?
FINALLY!!!!!!!!!!!!!!!!!!!
the ci test works for testing things. For the first time EVER!!!!!!!!!!!!!!!
Stuffs broken, but we have a working galaxy instance to test it against.

it is the testing...

I thought there was one, but I can't find it.

### DIFF
--- a/.github/workflows/ci_standalone.yml
+++ b/.github/workflows/ci_standalone.yml
@@ -49,9 +49,6 @@ jobs:
       - name: set the hub ui path in .compose.env
         run: echo "ANSIBLE_HUB_UI_PATH='/home/runner/work/ah_configuration/ah_configuration/ansible-hub-ui'" >> galaxy_ng/.compose.env
 
-      - name: disable approval setting override
-        run: sed -i.bak 's/PULP_GALAXY_REQUIRE_CONTENT_APPROVAL/#PULP_GALAXY_REQUIRE_CONTENT_APPROVAL/' galaxy_ng/dev/standalone/galaxy_ng.env
-
       - name: cat
         run: cat galaxy_ng/.compose.env
 
--- a/.github/workflows/ci_standalone.yml
+++ b/.github/workflows/ci_standalone.yml
@@ -1,0 +1,76 @@
+---
+name: Standalone
+on: 
+  pull_request:
+    branches:
+      - '*'
+    paths-ignore:
+      - 'docs/**'
+      - 'mkdocs.yml'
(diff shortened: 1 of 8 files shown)


sean-m-sullivan:
Note, this is pulled from here, they added some CI builds and there a few of them, I'm going to inquire about which does which that we can incorporate :)
https://github.com/ansible/galaxy_ng/blob/master/.github/workflows/ci_standalone.yml

djdanielsson (approved):
Even if it fails currently it is an improvement so I am approving it

.github/workflows/ci_standalone.yml
```diff
@@ -0,0 +1,97 @@
...
+        run: sed -i.bak 's/PIP_EDITABLE_INSTALL=1/PIP_EDITABLE_INSTALL=0/' galaxy_ng/.compose.env
+
+      - name: workaround github worker permissions issues
+        run: sed -i.bak 's/WITH_DEV_INSTALL=1/WITH_DEV_INSTALL=0/' galaxy_ng/.compose.env
+
+      - name: set the hub ui path in .compose.env
+        run: echo "ANSIBLE_HUB_UI_PATH='/home/runner/work/ah_configuration/ah_configuration/ansible-hub-ui'" >> galaxy_ng/.compose.env
+
+      - name: disable approval setting override
+        run: sed -i.bak 's/PULP_GALAXY_REQUIRE_CONTENT_APPROVAL/#PULP_GALAXY_REQUIRE_CONTENT_APPROVAL/' galaxy_ng/dev/standalone/galaxy_ng.env
```
Tompage1994: Should we not leave this in? Won't we want to test the functionality of approval as this is default behaviour anyway?

Tompage1994 (changes requested):
Happy with this as a start. There is one small question in here about the platform setup, but everything else looks good to go.

.github/workflows/ci_standalone.yml
djdanielsson: that is actually a good point, I do think we need to test the approval functionality because we have had reported issues recently about that

sean-m-sullivan: I'm thinking we should test both deployments, We can remove for now, but I'm thinking we should have both approval and non approval.

sean-m-sullivan: Done

Tompage1994 (approved):

Issue #135, opened 2022-08-01
ah_ee_repository fails when using include_tags


bprins:
When using `include_tags` with the `ah_ee_repository` module the Automation Hub API returns a HTTP 500 . From the `gunicorn` logging:
```
pulp [None]: django.request:ERROR: Internal Server Error: /api/galaxy/_ui/v1/execution-environments/remotes/
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/lib/python3.9/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/rest_framework/viewsets.py", line 125, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/lib/python3.9/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/usr/lib/python3.9/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/lib/python3.9/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/usr/lib/python3.9/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/usr/lib/python3.9/site-packages/rest_framework/mixins.py", line 19, in create
    self.perform_create(serializer)
  File "/usr/lib/python3.9/site-packages/rest_framework/mixins.py", line 24, in perform_create
    serializer.save()
  File "/usr/lib/python3.9/site-packages/rest_framework/serializers.py", line 212, in save
    self.instance = self.create(validated_data)
  File "/usr/lib64/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/usr/lib/python3.9/site-packages/galaxy_ng/app/api/ui/serializers/execution_environment.py", line 420, in create
    if "*-source" not in validated_data["exclude_tags"]:
TypeError: argument of type 'NoneType' is not iterable
```

I tried to set `exclude_tags` to an empty list but this is not allowed:
```
[...]

PR #143, opened 2022-08-31
Update to EE repository to fix labels and creation

sean-m-sullivan:
### What does this PR do?
Update the module to work, it was not working with labels as NONE was not allowed as an option for a list, so defaulted them to list. In addition a name has to be passed to create, so the 2nd time it was run it was not idempotent as no name was given if one existed.

Automated tests 

#135 I hope this fixes their issues, will check with them.


djdanielsson (approved):
LGTM

PR #146, opened 2022-09-02
Ready to Merge - Adding Dispatch Role

### LINKED ISSUE #145: Dispatch role like in controller_configuration
Currently, this collection requires that you call all of the roles for which you have variables. It would be nice if you could call just a single role that executes all the roles for which variables are found, like dispatch does in controller_configuration.

alawong:
### What does this PR do?
Added a dispatch role similar to that of controller_configuration's. Executes all roles for which variables are found.

TBC

Resolves #145 


Tompage1994 (commented):
Looks like a good start. Make sure you don't delete that galaxy.yml.j2 though as its used for CD

sean-m-sullivan:
This has some linting and is a WIP, I'm back and forth on the dispatch as it sometimes makes things TOO easy and complicated for new people to use, but once you have it set correctly, is an easy way to access everything. Thoughts?

alawong:
Noted Tom, this has been restored, I can't remember why I deleted this in the first place.

Both good points Sean. I'd argue that it doesn't make it more complicated, as all the role does is just run roles that variables exist for. With that understanding, I believe that even beginners should understand that the role just picks up the variables that they've written. 

Perhaps I can put in the readme that users should ensure that they're comfortable with the collection in general, before using the dispatch role? Do you think that would be enough?

Tompage1994:
> Perhaps I can put in the readme that users should ensure that they're comfortable with the collection in general, before using the dispatch role? Do you think that would be enough?

I think something about making sure to be familiar with the underlying roles would be a good idea. The dispatch role is a useful tool though in helping someone with some knowledge get a new controller server (or in this case an AH server) up and configured very quickly with minimal effort setting out the config as code.

silvinux (commented):
Hello,

The variable `ah_configuration_dispatcher_roles` should consider the order to create/modify objects because it may have dependencies. I suggest to use something like the below:

```
ah_configuration_dispatcher_roles:
  - {role: group, var: [ah_groups], tags: groups}
  - {role: user, var: [ah_users], tags: users} 
  - {role: ansible_config, var: [ansible_config_list, automation_hub_list], tags: config}
  - {role: namespace, var: [ah_ee_namespaces], tags: namespaces} 
  - {role: collection, var: [ah_collections], tags: collections}
  - {role: ee_repository, var: [ah_ee_repositories], tags: repos}
  - {role: ee_repository_sync, var: [ah_ee_repository_sync], tags: reposync}
  - {role: ee_image, var: [ah_ee_images], tags: images}
  - {role: ee_registry, var: [ah_ee_registries], tags: registries}
  - {role: ee_registry_index, var: [ah_ee_registries], tags: indices}
  - {role: ee_registry_sync, var: [ah_ee_registries], tags: regsync} 
```

Cheers,

Tompage1994 (approved):
This looks like its about there. @alawong could you retitle the PR out of WIP if its ready to go?

alawong:
@Tompage1994  done.

silvinux (approved):
LGTM

sean-m-sullivan (approved):
LGTM

PR #147, opened 2022-09-12
added ah_token to repository_sync role

gearboxscott:
### What does this PR do?
added the `ah_token` to the `repository_sync` role `tasks/main.yml`, that was missing.  When reading the example in the README.md for the role and in the `tasks/main` itself, it mentions using `ah_token`.  If you use `ah_token` it will not reload a new token in the Private Automation Hub for the `admin` or any other user account.  This will allow prevent reload of a token that will break job templates and projects in Ansible Automation Hub because the token was changed. 

We're just adding the missing variable that was omitted from the original code.  Adding the `ah_token:` to the `ah_auth.yml` in the `ah_config` directory with the current token should allow this role to run as intended originally per the original example.

n/a, just fixed it.

n/a

### DIFF
--- a/changelogs/fragments/collection.yml
+++ b/changelogs/fragments/collection.yml
@@ -1,4 +1,4 @@
 ---
 minor_changes:
-  - created a role from ah_collection module
+  - add `ah_token` so repository_sync works with token only and doesn't change the original token that happens when using `ah_username` and `ah_password`. This will keep it from breaking Ansible Automation Controller credentials that point to Private Automation Hub.
 ...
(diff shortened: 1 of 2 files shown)


changelogs/fragments/collection.yml
```diff
@@ -1,4 +1,4 @@
 ---
 minor_changes:
-  - created a role from ah_collection module
+  - add `ah_token` so repository_sync works with token only and doesn't change the original token that happens when using `ah_username` and `ah_password`. This will keep it from breaking Ansible Automation Controller credentials that point to Private Automation Hub.
```
djdanielsson: create new line do not delete someone else's change. also please shorten to:

add `ah_token` so repository_sync that was missing

djdanielsson (changes requested):
please fix changelogs and then I will approve

changelogs/fragments/collection.yml
gearboxscott: completed the requested changes

djdanielsson (approved):
small typo in changelog but that's ok, I found another issue I am working on a fix and I can correct it then. thanks for finding this issue and fixing it.

PR #148, opened 2022-09-14
added missing ah_token option

gearboxscott:
### What does this PR do?
adds the missing `ah_token` in all the roles that in the README.md but not in the role itself.

n/a

n/a

Related to `PR #147`.   Found more missing `ah_token` in the other roles that needed to be added so a new token is not generated when called.


djdanielsson (approved):
LGTM, thanks for finding these

Issue #139, opened 2022-08-05
Publish role is overwriting existing collection version


tonyreveal:
AAP 2.2.  
PAH Server version: 4.5.0
Pulp Ansible Version: 0.13.0

Trying to publish new version of existing collection with the `ah_overwrite_existing` var set to `false`.  However it always deletes the existing version.  I did see the note `This will only work if the account being used has correct privileges`.  We've tried the admin account with password and also tried a token.  We've also tried other super-user accounts with the same results.

When you run the playbook it will delete the collection and will fail with an error but you can see `overwrite_existing` is set to `true` (below output).  Then run the playbook again and it will 

```json
{
  "invocation": {
    "module_args": {
      "namespace": "cscdev",
      "name": "collectiontest",
      "path": "/var/tmp/Collectiontest/cscdev-collectiontest-1.1.13.tar.gz",
      "auto_approve": false,
      "overwrite_existing": true,
      "ah_host": "pah-dev.cscglobal.com",
      "ah_username": "admin",
      "ah_password": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
      "wait": true,
      "state": "present",
      "ah_path_prefix": null,
      "validate_certs": null,
      "ah_token": null,
      "version": null
    }
```

When you run the playbook again it will publish the new version of the collection.

djdanielsson:
where are you putting ah_overwrite_existing to false? can you print out the hostvars at the start of the playbook to see if you see it as true or false?

tonyreveal:
They originally had it as a parameter in the collection defintion.  I changed it to pass it as a var to the role which solved the `overwrite_existing: true` issue.  However, now it doesn't do anything.  It keeps the existing version and does not publish another version (isn't keeping old versions).  The expectation is that they be able to click the Version drop down box and see many versions to choose from.

```yaml
- name: Import ah_configuration role
  import_role:
    name: redhat_cop.ah_configuration.publish
  vars:
    ah_overwrite_existing: false
```

djdanielsson:
thanks for the update, I agree I would expect that result. We will have to try and do some deep debugging but that might take a while.

Tompage1994:
What relevant vars are set for this role? Is a version set in the galaxy.yml? The version doesn't get bumped by this role, so if an existing collection exists of the matching version and you have told it not to overwrite existing it should do nothing in theory

tonyreveal:
[...]

PR #149, opened 2022-09-15
fixing #139

djdanielsson:
### What does this PR do?
This should solve #139 

manual

#139

none


sean-m-sullivan (approved):
LGTM, this should be that.

PR #154, opened 2022-09-27
Set Global proxy config at ee_registry tasks

silvinux:
### What does this PR do?

Proxy variables at task roles/ee_registry/tasks/main.yml are set to be loaded from a list instead to use a global variable. To be consistent with other roles that these variables are used (roles/repository/tasks/main.yml) and as it should be a global variable I've changed.   

Configure ee-registries with proxy variables.


djdanielsson (changes requested):
I think this is probably fine but you need to update the readme to match this change

silvinux:
> eadme to match this change
I've edited the role's README.md to match the change as requested.

djdanielsson (approved):
LGTM

PR #162, opened 2022-10-06
Draft: add ah token to auth ui doc fragment

sean-m-sullivan:
### What does this PR do?

This should add the ah_token to multiple modules that are currently missing it. 

modules include this file like so
https://github.com/redhat-cop/ah_configuration/blob/devel/plugins/modules/ah_user.py#L79

### DIFF
--- a/plugins/doc_fragments/auth_ui.py
+++ b/plugins/doc_fragments/auth_ui.py
@@ -31,6 +31,14 @@ options:
     - Password for your Ansible Galaxy or Automation Hub instance.
     - If value not set, will try environment variable C(AH_PASSWORD)
     type: str
+  ah_token:
+    description:
+    - The Ansible Galaxy or Automation Hub API token to use.
+    - This value can be in one of two formats.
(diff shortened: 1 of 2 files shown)


djdanielsson (approved):
hopefully this will fix #163

ivarmu (approved):
LGTM

silvinux (approved):
LGTM. Nice catch!

djdanielsson (changes requested):
waiting for confirmation that this does fix the issue before approving and merging

ivarmu:
This is (by now) only a documentation fragment addition... this won't solve nothing at all:

image

sean-m-sullivan:
This needs more so WIP

ivarmu:
Just to give some information:

When authenticating with `username` and `password`, the endpoint to be used seems to be `/api/galaxy/_ui/v1/auth/login/`.
When authenticating with `token`, the enpoint to be used seems to be `/api/galaxy/v3/auth/token/`.

djdanielsson:
@sean-m-sullivan should we close this?

PR #174, opened 2022-12-06
update workflows

sean-m-sullivan:
# What does this PR do?
update workflows


sean-m-sullivan:
@Tompage1994 I am going to need help on this one.

Tompage1994:
Love this. So much testing going on. Works especially well here with the modules. 

I think I've fixed all the Python stuff now and pushed to your branch. Should be good 🤞

Tompage1994:
Ok, so issues lies with the fact the collection is built for the sanity check as infra.ah_configuration but the doc_fragment is set as redhat_cop.ah_configuration.

This is symptomatic of a problem we have with releasing modules here. The doc_fragments only work if the name of the collection is hardcoded in. We need to make sure the the below line in ALL the modules and plugins is replaced with infra for the infra.ah_configuration collection.

```
extends_documentation_fragment: redhat-cop.ah_configuration.auth
```

Tompage1994:
Created #175 for this issue

Tompage1994:
Ok, I've updated all references to redhat_cop.ah_configuration to infra.ah_configuration to make this work.

I don't know whether we want to go with that though. It will mean that we have to fix the doc_fragments for releases to redhat_cop namespace

Tompage1994:
I also removed Black as the Python linter since the ansible-test uses pep8 and the action didn't like ignoring it.

Tompage1994:
ok, lots of action but this should be good now. To summarise the key things that actually change here:

- changed all references of `redhat_cop.ah_configuration` to `infra.ah_configuration`
- Removed Black as a linter

I've approved this PR, but now given I've done a lot of changes too we probably need a third pair of eyes

Tompage1994 (approved):

djdanielsson:
The workflow can sed the file to fix it for the 6 months we support both namespaces.... yes it's a hack but it should work

sean-m-sullivan:
We can add that to the workflow yes. Was going to suggest we change the variable, but Tom's made it work. I want to run tests on things too, just in case any of the lint changes broke something in the python.

changelogs/fragments/linting.yml
```diff
@@ -0,0 +1,5 @@
+---
+bugfixes:
+  - Fixed doc issue in ah_collection role where overwrite_existing option was reported with default True when functionality was False
+  - Fixed doc issue in ah_collection_upload role where wait option was reported with default True when functionality was False
+...
```
djdanielsson, line 13: also fixed issue where some private things were not set to no_log, that's fairly big and worth noting.

djdanielsson (changes requested):
this seems like a really big update, though I know a lot of it is just renaming or changing format. I didn't notice anything jump out at me except for changelog not talking about adding no_log to a few vars which I think is valuable.

changelogs/fragments/linting.yml
Tompage1994: Your right, I forgot to continue adding these changelogs for all the things I fixed. I'll add those now

Tompage1994:
Updated with changelogs and a couple of fixes to docs in roles and a minor change to the role behaviour which didnt match the module

djdanielsson (approved):
LGTM

sean-m-sullivan (commented):
LGTM, have given it a few passes over the morning.

Tompage1994:
Opened #176 for outstanding CI failure

Issue #127, opened 2022-07-08
redhat_cop.ah_configuration.ah_approval fails with error like [error] Bash exited with code 2


gabhishek627:
- redhat_cop.ah_configuration (v 0.8.0)
- The task for redhat_cop.ah_configuration.ah_collection works fine
- while approving the collection ( via redhat_cop.ah_configuration.ah_approval) it fails with error like [error] Bash exited with code 2
-  The failure with the approval appears randomly for any collection

- Approve collections with retries to 10
TASK [Approve Collections] *****************************************************
FAILED - RETRYING: [localhost]: Approve Collections (10 retries left).
FAILED - RETRYING: [localhost]: Approve Collections (9 retries left).
FAILED - RETRYING: [localhost]: Approve Collections (8 retries left).
FAILED - RETRYING: [localhost]: Approve Collections (7 retries left).
FAILED - RETRYING: [localhost]: Approve Collections (6 retries left).
FAILED - RETRYING: [localhost]: Approve Collections (5 retries left).
FAILED - RETRYING: [localhost]: Approve Collections (4 retries left).
FAILED - RETRYING: [localhost]: Approve Collections (3 retries left).
FAILED - RETRYING: [localhost]: Approve Collections (2 retries left).
FAILED - RETRYING: [localhost]: Approve Collections (1 retries left).

Tompage1994:
Could you show what the error looks like after all the retries have failed? Could you run your play with `-vvv` and provide the full output for this task please?

Tompage1994:
Could you also ensure there is a collection awaiting approval in the UI?

gabhishek627:
No, Whenever it fails the approval never appears, the failed ones are still unavailable in the approval screen.

gabhishek627:
TASK [Approve Collections] *****************************************************
FAILED - RETRYING: [localhost]: Approve Collections (10 retries left).
FAILED - RETRYING: [localhost]: Approve Collections (9 retries left).
FAILED - RETRYING: [localhost]: Approve Collections (8 retries left).
FAILED - RETRYING: [localhost]: Approve Collections (7 retries left).
FAILED - RETRYING: [localhost]: Approve Collections (6 retries left).
FAILED - RETRYING: [localhost]: Approve Collections (5 retries left).
FAILED - RETRYING: [localhost]: Approve Collections (4 retries left).
FAILED - RETRYING: [localhost]: Approve Collections (3 retries left).
[...]

PR #192, opened 2023-01-02
Updates to ah_collection

sean-m-sullivan:
# What does this PR do?
Updates to ah_collection

# How should this be tested?
Automated tests

After review, and looking at tasks, the ah_collection auto approval Absolutely needs the version, either from the tarball, or from creation, so this will grab it if the user doesn't provide it. This should eliminate errors with the module where it could not approve and would time out.


djdanielsson (approved):
LGTM

sean-m-sullivan:
Linking issue #127

PR #195, opened 2023-01-08
Update Testing

sean-m-sullivan:
# What does this PR do?
Updating testing to add more coverage to fix approval testing for 4.5/4.4
Add back publish role testing, fix publish role to be namespace agnostic like other roles.


roles/publish/tasks/main.yml
```diff
@@ -29,7 +29,7 @@
     - __ah_collection_item.collection_local_path is defined
 
 - name: Build Collections
-  infra.ah_configuration.ah_build:
```
djdanielsson, line 32: I am fine with this until 2.4 then we should FQCN everything possible to infra

.github/tests/playbooks/testing_playbook.yml
```diff
@@ -36,20 +36,15 @@
...
     - name: Upload Collection
       ansible.builtin.include_role:
         name: collection
 
-    #  - name: Publish Collections
-    #    ansible.builtin.include_role:
-    #      name: publish
-    #    vars:
-    #      ah_auto_approve: true
+    - name: Publish Collections
```
djdanielsson, line 43: Shouldn't this say publish? I'm on mobile so kinda hard to tell 100%

djdanielsson (approved):
LGTM

roles/publish/tasks/main.yml
sean-m-sullivan: When we switch it over we should switch the testing over to use infra as well, and all the roles at the same time, Agreed, This was more lets make it congruent in the meantime, and was causing testing fail as we were building with redhat_cop.

.github/tests/playbooks/testing_playbook.yml
sean-m-sullivan: it does say publish, thought this was another.

PR #197, opened 2023-01-09
Fix issue where user thought they could add token to ah_api lookup plugin

Tompage1994:
# What does this PR do?
Remove unused oauth2 option from documentation and ensure username and password are provided to ah_api lookup

The doc previously said you could apply a token to the plugin but you cant. I've removed this and added a check if username and password are set and errors out with a better error message if not.

# How should this be tested?
```
    - ansible.builtin.debug:
        msg: "Collections: {{ lookup('infra.ah_configuration.ah_api', 'namespaces', host=ah_hostname, username=ah_username, password=ah_password, validate_certs=false, return_all=true) }}"
```
will work, but 
```
    - ansible.builtin.debug:
        msg: "Collections: {{ lookup('infra.ah_configuration.ah_api', 'namespaces', host=ah_hostname, validate_certs=false, return_all=true) }}"
```
will not work.

# Is there a relevant Issue open for this?

resolves #196 

# Other Relevant info, PRs, etc

N/A

Issue #196, opened 2023-01-08
Lookup plugin ah_api is not working


Alex-Bron:
# Summary

When using the ah_api lookup plugin, tasks fail with message:
`unknown doc_fragment(s) in file /home/alex/git/aap/collections/ansible_collections/redhat_cop/ah_configuration/plugins/lookup/ah_api.py: infra.ah_configuration.auth_plugin
`
When manually replacing infra.ah_configuration with redhat_cop.ah_configuration, tasksfail with message:
`An unhandled exception occurred while running the lookup plugin 'redhat_cop.ah_configuration.ah_api'. Error was a <class 'AttributeError'>, original message: 'dict' object has no attribute 'getheaders'. 'dict' object has no attribute 'getheaders'`

# Issue Type

- Bug Report

# Ansible, Collection, Private Automation Hub details

```console (paste below)
ansible --version
ansible [core 2.14.1]
  config file = /home/alex/git/aap/ansible.cfg
  configured module search path = ['/home/alex/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  ansible collection location = /home/alex/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.9.13 (main, Nov  9 2022, 13:16:24) [GCC 8.5.0 20210514 (Red Hat 8.5.0-15)] (/usr/bin/python3.9)
  jinja version = 3.0.3
  libyaml = True

ansible-galaxy collection list
# /usr/share/ansible/collections/ansible_collections
Collection       Version
---------------- -------
redhat.rhel_mgmt 1.1.0  

# /home/alex/.ansible/collections/ansible_collections
Collection              Version
----------------------- -------
ansible.posix           1.4.0  
community.crypto        2.7.1  
community.general       5.8.0  
community.libvirt       1.2.0  
containers.podman       1.9.4  
freeipa.ansible_freeipa 1.8.4  
oasis_roles.system      1.3.0  
redhat.insights         1.0.7  
redhat.rhel_idm         1.8.4

Private Automation Hub version
2.3
```

- ansible installation method: one of source, pip, OS package, EE
OS package and EE

# OS / ENVIRONMENT

Tasks run on localhost (RHEL8) against Private Automation Hub 2.3 on RHEL 8.7

# Desired Behavior

[...]

PR #199, opened 2023-01-09
Update collections api lookup to work across all repos

### LINKED ISSUE #198: ah_api lookup doesn't bring all collections when search term is 'collections'
The lookup plugin does not work to pull down collections from all repositories. This may only be an issue with AAP2.3.

Suggested solution is to default to pulling published collections (as currently does) but also allow an additional term to specify the repository.

>         Thanks for the very swift replies to my issue. So I changed the namespace to infra (will redhat_cop.controller_configuration also be replaced by infra.controller_configuration?) and added my username and password to the ah_api lookup, however this returns an empty list although I'm sure there are collections present on my automation hub.
> Am I maybe doing something wrong here? Is my assumption incorrect that this ah_api lookup can be used to retrieve a list of collections present on the system? In the end I'd want to query for a list of collections from the rh-certified repository, as those should not be searched for in the community repository.
> 
> Code snippet:
> ```
> - name: Next, query the Automation Hub for the rh-certified collections
>   ansible.builtin.debug:
>     msg: "Collections: {{ lookup('infra.ah_configuration.ah_api', 'collections', host=ah_hostname, username=ah_username, password=ah_password, return_all=true) }}"
> ```
> 

_Originally posted by @Alex-Bron in https://github.com/redhat-cop/ah_configuration/issues/196#issuecomment-1375590074_

Tompage1994:
# What does this PR do?

Updates the lookup plugin for the API to work with the collections query to pull from a specific repository. It seems the URL has changed (although redirected) which meant that only the published collections were being pulled. There doesn't seem to be a way to get all the collections now so sensible thing to do is to take the repository we want to query and pull from that.

# How should this be tested?
```
- ansible.builtin.debug:
        msg: "Collections: {{ lookup('infra.ah_configuration.ah_api', 'collections', 'community', host=ah_hostname, username=ah_username, password=ah_password, validate_certs=false, return_all=true) }}"
```

also if the repository is not provided it shall still work and default to 'published'

resolves #198 

related #196 
related #197

### DIFF
--- a/changelogs/fragments/api_lookup.yml
+++ b/changelogs/fragments/api_lookup.yml
@@ -1,0 +1,7 @@
+---
+minor_changes:
+  - Added option to pull all collections from a specific repository
+bugfixes:
+  - Checks for username and password passed to api lookup plugin
+  - Removed incorrect reference to oauth token abilities for api lookup plugin.
+...
--- a/plugins/lookup/ah_api.py
+++ b/plugins/lookup/ah_api.py
@@ -21,8 +21,8 @@ options:
       - "'ee_namespaces'"
       - "'ee_registries'"
       - "'ee_repositories'"
-      - "'collections'"
-      - "'collection', [repository={published, rh_certified, community}], [collection_namespace], [collection_name]"
+      - "'collections', [repository={published, rh_certified, validated, community}]"
+      - "'collection', [repository={published, rh_certified, validated, community}], [collection_namespace], [collection_name]"
       - "'groups'"
       - "'namespaces'"
       - "'repository_community'"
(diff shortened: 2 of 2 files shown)


changelogs/fragments/api_lookup.yml
```diff
@@ -0,0 +1,7 @@
+---
+minor_changes:
+  - Added option to pull all collections from a specific repository
+bugfixes:
```
Tompage1994, line 4: Note that these are from #197 but I forgot the changelog and its already been merged.

branic: and I missed that there was not changelog fragment. :facepalm:  Thanks for catching that @Tompage1994

sean-m-sullivan (approved):
LGTM

PR #201, opened 2023-01-12
add Ah role and other improvements

sean-m-sullivan:
# What does this PR do?
add role module to add group roles. This will not be able to be assigned to group permissions until a bug is fixed in galaxy ng.

In addition some endpoints were moved in latest version
https://github.com/ansible/galaxy_ng/pull/1397/files

after testing all the changes beyond the module should be backward compatible, but do not fix any of the underlying issues with ee_namespace or repository modules that still need to be updated.

# How should this be tested?
Automated tests


plugins/modules/ah_role.py
```diff
@@ -0,0 +1,223 @@
...
+__metaclass__ = type
+
+
+DOCUMENTATION = r"""
+---
+module: ah_role
+short_description: Manage a role of group permissions
+description:
+  - Manage a role of group permisions
+version_added: '1.0.0'
```
Tompage1994: This should probably be 1.1.0 because we are already on 1.0.1 and we will probably do a minor release to push this change

Tompage1994 (changes requested):
Very minor change requested but LGTM.

I like that you've used the get_server_version. Something I had not thought of before (duh) and probably could be used more widely on some of the issues we have found with changing APIs

plugins/modules/ah_role.py
sean-m-sullivan: Oh I tried that, and our choices are 1.0.0 or 1.1.0, the sanity tests will not allow module additions in minor releases

Tompage1994 (approved):

Issue #184, opened 2022-12-08
Workflow for ee_repositories does not work


Tompage1994:
# Summary
The workflow for EE Repos does not work. The issue is around adding in an image and there's a definite issue with doing the podman push. Perhaps we can think of a better way to test all this out.

# Issue Type

- Bug Report

# OS / ENVIRONMENT
CI

# Actual Behavior

```console (error)
TASK [Ensure the image is pushed in private automation hub] ********************
fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["podman", "push", "--tls-verify=false", "--remove-signatures", "--creds=admin:redhat", "localhost:8002/test-namespace/ee-minimal-rhel8:latest"], "delta": "0:00:00.079395", "end": "2022-12-07 16:11:32.954052", "msg": "non-zero return code", "rc": 125, "start": "2022-12-07 16:11:32.874657", "stderr": "Error: trying to reuse blob sha256:50644c29ef5a27c9a40c393a73ece2479de78325cae7d762ef3cdc19bf42dd0a at destination: pinging container registry localhost:8002: invalid status code from registry 404 (Not Found)", "stderr_lines": ["Error: trying to reuse blob sha256:50644c29ef5a27c9a40c393a73ece2479de78325cae7d762ef3cdc19bf42dd0a at destination: pinging container registry localhost:8002: invalid status code from registry 404 (Not Found)"], "stdout": "", "stdout_lines": []}
```

sean-m-sullivan:
With the recent merges and from what I saw this should be fixed. It might be around the renaming issue. Closing unless Tom wants to reopen.

Issue #183, opened 2022-12-08
Update EE namespace for AH 4.6


Tompage1994:
# Summary

The ee_namespace API has changed in 4.6 so the module is broken there. 

There is now a _ui API so this can be redesigned. I don't know whether there was in 4.5 or before. We really don't need to be using the pulp API for this because its a PITA.

`https://ah.aap23.local/api/galaxy/_ui/v1/execution-environments/namespaces/ee-29-rhel8/ `

# Issue Type

- Bug Report

sean-m-sullivan:
It looks like in 4.6 The namespace API was discontinued to just be repositories. I'll be putting in a PR to note that in the module and role.

https://github.com/ansible/galaxy_ng/pull/1397

sean-m-sullivan:
Closing this with the recent merge, if we want to revisit, we can reopen, but I don't think the API is changing back.

PR #204, opened 2023-01-17
Namespace and registery update

sean-m-sullivan:
# What does this PR do?
Updates registry and namespace to work with newer versions
Reinstates registry test back into regular tests

# How should this be tested?
Automated tests

#184 , #183 Maybe more, need to add and test around remote repos still.


plugins/module_utils/ah_ui_object.py
```diff
@@ -1043,6 +1043,89 @@ def __init__(self, API_object, data=None):
...
+            # Only one object should be returned. If more that one is returned,
+            # then look for the requested name in the returned list.
+            for asset in response["json"]["data"]:
+                if self.name_field in asset and asset[self.name_field] == name:
+                    self.data = asset
+                    self.exists = True
+                    return
+            self.data = {}
+            self.exists = False
+            return
```
branic, line 1121: This seems like it could return different results across calls as the order of the assets returned from AH may not always be in the same order. I think it would be better in this case to raise an error. Or am I missing something?

sean-m-sullivan: This is more of a catch all in case we have similar named objects returned, its a duplicate to what we've been using the UI parent function, and most of the time it should not trigger. This should all go away in the future as the UI API endpoint is slowly being depreciated and retired.

branic: That makes sense. Then this LGTM

branic (approved):
LGTM

PR #207, opened 2023-01-19
Add sync and index option on respective roles

### LINKED ISSUE #167: ah_ee_registry_index failing for custom EE remote registries
When configuring a custom EE remote registry (Quay for instance), the ee_registry_index task fails because indexing is only supported for registry.redhat.io

There should be a variable like "index: true|false" for ah_ee_registries and ah_registry_index should not try indexing when set to false.

Tompage1994:
# What does this PR do?
Adds sync or index option on roles to conditionally sync or index. This is based around the fact that you can't index all remote repositories so if you call the index role it will cause issues if one of your repositories isn't registry.redhat.io.

I figure it makes sense to do the same thing with the sync role too since the user may not want to sync either.

# How should this be tested?
Not sure how great it would be to add to CI for _index and _sync for EEs but it can be tested by calling the role with or without sync and index set as true as options of `ah_ee_registries`

resolves #167 

### DIFF
--- a/changelogs/fragments/sync-index.yml
+++ b/changelogs/fragments/sync-index.yml
@@ -1,0 +1,5 @@
+---
+breaking_changes:
+  - Introduces a `sync` option to the ee_registry_sync role on the `ah_ee_registries` variable which by default is false and which is required true to sync the registry.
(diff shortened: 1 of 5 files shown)


sean-m-sullivan (approved):
LGTM this is a littel counter intuitive, but does solve how you keep them all in a list rather then calling them individually as roles.

Issue #182, opened 2022-12-07
Error when creating new EE registry


spyrexd:
# Summary

The ee_registry role fails with `"AttributeError: 'dict' object has no attribute 'getheaders'"`. This happens after succesfully configuring namespaces

# Issue Type

- Bug Report

# Ansible, Collection, Private Automation Hub details

```console (paste below)
ansible --version

ansible [core 2.14.0]
  config file = /home/mataylor/.ansible.cfg
  configured module search path = ['/home/mataylor/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/mataylor/venv/ansible/lib64/python3.10/site-packages/ansible
  ansible collection location = /home/mataylor/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/mataylor/venv/ansible/bin/ansible
  python version = 3.10.8 (main, Nov 14 2022, 00:00:00) [GCC 12.2.1 20220819 (Red Hat 12.2.1-2)] (/home/mataylor/venv/ansible/bin/python3)
  jinja version = 3.1.2
  libyaml = True

ansible-galaxy collection list

redhat_cop.ah_configuration         0.9.2  
redhat_cop.controller_configuration 2.1.6  

Private Automation Hub version

4.4.0

```

- ansible installation method: pip

# OS / ENVIRONMENT

Custom EE built on egistry.redhat.io/ansible-automation-platform-22/ee-supported-rhel8:latest running as an image on a gitlab runner

# Desired Behavior

EE Registry is created in PAH

# Actual Behavior

Playbook executes and is able to configure a namespace then fails when creating the registry.

Main error
```
"/tmp/ansible_ah_ee_registry_payload_uo7x_ta6/ansible_ah_ee_registry_payload.zip/ansible_collections/redhat_cop/ah_configuration/plugins/module_utils/ah_api_module.py\", line 371, in authenticate", "AttributeError: 'dict' object has no attribute 'getheaders'"
```

Console output of error
```console (error)
TASK [redhat_cop.ah_configuration.namespace : Add Automation Hub Namespace] ****
ok: [localhost] => (item={'name': 'REDACTED', 'groups': [{'name': 'REDACTED', 'object_permissions': ['change_namespace', 'upload_to_namespace']}]})
TASK [redhat_cop.ah_configuration.namespace : Create Namespace | Wait for finish the namespace creation] ***
[...]

Issue #138, opened 2022-08-02
current code requires 2.2


djdanielsson:
when doing some testing I ran into an issue about some of the new content we wrote and it errors because I was not running the latest version of AAP. This should probably be something we track a little better and note in releases when changes require specific versions.

sean-m-sullivan:
We've added a version to get and track and change based on what version the user is pointing at. We should continue doing this, but closing this issue as solved.

PR #208, opened 2023-01-22
Update tests, fix many things, and add role role.

sean-m-sullivan:
# What does this PR do?
This adds 83% test coverage to every option feasible to test, making sure all of those are done with at least a role where possible. There are about 22/125 options we are unable to test due to proxy, certs, and other things, but this brings up the testing considerably, and fixes problems where found.

minor_changes:
  - Updated ah_collection_upload to exit properly.
  - Updated testing coverage.
  - Updated ah_role to not require permisions in order to allow for deletion.
  - Added ansible role 'role' for creating roles of group permsions.
  - Fixed issue with ah_ee_repository working on newer versions.
  - Fixed issue with ah_namespace delete
  - ah_ee_repository include/exclude tags are no longer mutually exclusive
  - added better error catching to ah_ee_registery_index/sync to handle when a registery is not found.
breaking_changes:
  - Introduces a `sync` option to the ee_repository_sync role on the `ah_ee_repositories` variable which by default is false and which is required true to sync the repository.
  - Removed ah_ee_registry new name option. This affected underlying images indexed from the registery leaving them abandoned, and subsequent indexes of the renamned registery would fail.

# How should this be tested?
This should be tested with the testing that we have in addition to the testing added.

#184 
#183 
#182 
#138
#135


sean-m-sullivan:
Because of the breadth of the the changes, I want Tom to review this before we merge it.

sean-m-sullivan:
Also because of added variables aren't allowed in PR's, it will fail, however here is the same commit passing: https://github.com/sean-m-sullivan/ah_configuration/actions/runs/3977449123

silvinux (commented):
Awesome job sean!!!. Looks awesome!

plugins/modules/ah_role.py
```diff
@@ -159,7 +158,7 @@ def main():
     argument_spec = dict(
         name=dict(required=True),
         description=dict(),
-        perms=dict(type="list", elements="str", required=True),
+        perms=dict(type="list", elements="str"),
```
Tompage1994, line 161: If there is no required here, do we not want to default this to `[]`?

Tompage1994 (approved):
LGTM - Though theres one question in there that I'm not sure about, happy if you don't think it needs to be fixed though

plugins/modules/ah_role.py
Tompage1994: And upon checking, yes, this does need to be defaulted to `[]` otherwise it fails

sean-m-sullivan: Need to do a test, but going to fix it.

sean-m-sullivan: ok fixed and tested, also found another error, but for now its just only set the perms defined, we could add some fancy add to, etc, but made it static like the roles in awx.awx.

PR #227, opened 2023-02-16
Skip ansible-lint rules

Tompage1994:
# What does this PR do?
Fixes lint issues caused by update to ansible-lint 6.13.0

# How should this be tested?
CI should pass

# Is there a relevant Issue open for this?

N/A

# Other Relevant info, PRs, etc
Note that I've set jinja[invalid] to warn for now because the new version has caused a regression in this rule which I've raised a bug for and linked to from the PR in controller_configuration https://github.com/redhat-cop/controller_configuration/pull/503

PR #231, opened 2023-02-21
Ci fix

### LINKED ISSUE #230: Roles and modules test doesn't run on PR
# Summary
I've been wondering for ages why the CI is failing for all PRs in this repo. It turns out the secrets aren't being passed to the jobs correctly.

# Issue Type

- Bug Report

Tompage1994:
# What does this PR do?
Allows secrets to be passed to CI to allow PRs to successfully be tested.

# How should this be tested?
PR CI will work. You may not be able to see it until merged though

resolves #230 

[This article](https://securitylab.github.com/research/github-actions-untrusted-input/) explains the usage of pull_request_target instead of pull_request. Note that it does add security implications if you run from users you don't trust, but GH requires maintainers to allow actions to be run on the first PR.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -14,8 +14,12 @@ use_default_rules: true
 # with verbosity set to 1, its dumping 'unknown file type messages'
 # verbosity: 1
 skip_list:
-  - meta-runtime
+  - meta-unsupported-ansible
+  - meta-runtime  # This collection with the appropriate awx.awx or ansible.controller still works with older ansible.
   - fqcn[keyword]
+  - role-name[path]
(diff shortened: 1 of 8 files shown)


Tompage1994:
Pre-commit will fail because of #227  waiting to be merged. Merge that first

Tompage1994:
The way the pull_request_target works is that it runs the workflow from `devel` so this PR clearly means the CI wont run. So it's a case of needing to merge this and raise a dummy PR to test

sean-m-sullivan (approved):
LGTM, looks like it will work

PR #233, opened 2023-02-22
Test PR <DO NOT MERGE>

Tompage1994:
# What does this PR do?
Tests whether updated CI works

Expecting the CI for testing roles and modules to correctly work here


Tompage1994:
Confirmed that where the PR CIs were previously failing is now getting past (i.e. the secrets are being propagated to the playbook)

There is a different CI failure now which I'll raise a separate bug for

PR #238, opened 2023-03-09
add missing ah_token option for community repository

dbk-rabel:
# What does this PR do?
Commit 5e5f76d did add ah_token for the certified repository, but not for community repository. This commit adds that.

# How should this be tested?

Use the simple example from https://github.com/redhat-cop/ah_configuration/blob/devel/roles/repository/README.md with ah_token as authentication method.

Didn't find one.

-


dbk-rabel:
By the way: I have no idea, why the automatic tests could be failing on this :-(

Tompage1994 (approved):
Good catch, thanks for the contribution

Tompage1994:
CI broken but inconsequential of this. Will merge this and raise a separate issue to fix CI

PR #243, opened 2023-03-20
Many changes to ansible_config role

Tompage1994:
# What does this PR do?
I made a start with the arg_spec for this collection and immediately ran into the ansible_config role which had many issues. This should fix them all. There are some real changes here rather than just doc fixes so this should be thoroughly checked. 

I suspect that this role doesn't get used that often as I'm not sure it really would have worked before.

# How should this be tested?
Added some CI for this

N/A

resolves #[number]

N/A


roles/ansible_config/templates/ansible.cfg.j2
```diff
@@ -1,19 +1,19 @@
...
+{% if server.auth_url is defined %}
+auth_url={{ server.auth_url }}
 {% endif %}
 {% if server.token is defined %}
 token={{ server.token }}
-{% endif %}
-{% if ah_token.token is defined %}
-token={{ ah_token.token }}
+{% elif ah_token is defined %}
+token={{ ah_token }}
```
djdanielsson, line 17: there is no else here so it will error if neither server.token nor ah_token is defined. I wonder if it is more simple to just do {{ server.token | default(ah_token) }} ? because unless I am still not fully awake the result is the same.

djdanielsson (approved):
I have one thought but it isn't big enough to request a change if what I said doesn't make sense we can just merge this.

changelogs/fragments/ansible_config
```diff
@@ -0,0 +1,9 @@
+---
```
djdanielsson: it looks like the CI won't pass if this isn't .yml file

djdanielsson (changes requested):
actually after double checking tests it looks like there is a change that needs to happen

roles/ansible_config/templates/ansible.cfg.j2
Tompage1994: Won't be an error, will just fall out of the conditional and not do anything. The default would work as well but I think I prefer this as it's a little easier to tell whats going on

djdanielsson (approved):
LGTM

djdanielsson:
ci_standalone master seems stuck so I will force merge

Tompage1994:
> ci_standalone master seems stuck so I will force merge

Ah yes, that was removed and still set as a dependency for merge on main. Have removed as a dependency in the repo settings

PR #245, opened 2023-03-20
Add arg specs

Tompage1994:
# What does this PR do?
Add arg spec for all roles

# How should this be tested?
CI

N/A

related https://github.com/redhat-cop/controller_configuration/discussions/509


djdanielsson (changes requested):
tests are failing

Tompage1994:
> tests are failing

I'll probably have to work through any issues so don't merge whilst this is failing, but once it's passed then we should be good to go. I suspect there will be a few bits where the argspec causes issues here so I'll need to resolve

Tompage1994:
Realising now that changes to CI don't get reflected in PRs anymore because of the workflow. I've fixed the first issue but possibly worth approving this and I'll fix anything broken tomorrow

Tompage1994:
The CI will never work because it's currently broken in main and the new way of running CI means changes to .github aren't used (i.e. it takes the pipeline code from main to run CI)

Tompage1994:
Have raised #246 to fix CI not running against updated test cases

PR #246, opened 2023-03-22
Move tests into codebase so updates run on CI

Tompage1994:
# What does this PR do?
Since we changed the CI to run on 'pull_request_target', the CI only runs against the tests set on devel, but we should probably be able to update the tests and them be run. This change moves them into the tests directory which is probably more appropriate

# How should this be tested?
CI will run after its been merged (not ideal, i know but this should resolve that)

No

The CI kind of works here: https://github.com/Tompage1994/ah_configuration/actions/runs/4489326679/jobs/7895008301 (i.e. you can see it runs in the new location, though I anticipate it will fail as I don't have secrets configured on my personal repo fork)


djdanielsson (approved):
LGTM

PR #262, opened 2023-06-26
Update .ansible-lint to ignore var naming

sean-m-sullivan:
# What does this PR do?
Fix linting, right now we do use unique vars, but her and controller_config we added this ignore.


djdanielsson (approved):
LGTM

PR #271, opened 2023-07-18
Fix repository roles

sean-m-sullivan:
# What does this PR do?

Makes fixes to repository roles. 
Also renamed the sync/delay variable to be less of a mouthful as plural, since we are reintroducing this, should not break anything.

# How should this be tested?
Automated testing, the main branch is breaking atm.


djdanielsson (approved):
thanks

PR #273, opened 2023-07-20
add ah_configuration_async_dir var

benformosa:
Allows setting ansible_async_dir, to work around issues with the value used by the collection.

# What does this PR do?

Creates a new variable `ah_configuration_async_dir` on all roles which use the async keyword on tasks, which sets the `ansible_async_dir` variable on those tasks.

I'm trying to use these roles from Ansible Automation Platform, and I get the error "could not find job" from the `ansible.builtin.async_status` tasks.
I'm currently working with Red Hat support to troubleshoot the issue.
In the mean time, I propose updating this collection to allow setting the ansible_async_dir to Ansible's default, or some other directory.
The issue I'm having seems to be related to redhat-cop/controller_configuration issue 299

This change could probably be made much simpler by removing the var from each role. That would break compatibility, but anyone wanting to restore that behaviour could set `ansible_async_dir` in their playbook.

It was also not clear how to handle a variable that was shared by all roles. I've done my best to imitate how some of the other variables were defined.
Something I noticed was the way that variables are described in role READMEs are a bit inconsistent, compare roles/ee_namespace/README.md and roles/repository/README.md.

# How should this be tested?

I haven't updated tests here, but I have successfully tested this change with the `infra.ah_configuration.namespace` role on Ansible Automation Platform.

I guess an automated test could run a role three times, with `ah_configuration_async_dir` unset, set to some path, and set to null.

### DIFF
--- a/roles/collection/README.md
+++ b/roles/collection/README.md
@@ -20,6 +20,10 @@ These are the sub options for the vars `ah_collections` which are dictionaries w
 |`overwrite_existing`|"false"|no|Overwrites an existing collection and requires version to be set.||
 |`state`|"present"|no|Desired state of the resource||
 
+The `ah_configuration_async_dir` variable sets the directory to write the results file for async tasks.
+The default value is `/tmp/.ansible_async`.
+Set to `null` to use Ansible's default value.
+
 ### Secure Logging Variables
 
 The following Variables compliment each other.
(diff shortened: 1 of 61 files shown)


sean-m-sullivan:
This broke some tests, and we want to look a little bit deeper and double check how this can impact things. 

However this is a good PR and was not an issue that we were aware of, and the PR overall is well done. It just may require a tweak or two in the end. 

Thanks for your submission, and we'll be looking at this more shortly.

sean-m-sullivan:
After testing, going to merge, we are debating whether to default to null, and fixing the testing errors

sean-m-sullivan (approved):

PR #276, opened 2023-07-24
Set Async dir to Ansible Default

sean-m-sullivan:
# What does this PR do?
This PR fixes two tests and sets the async dir to the ansible default. We used tmp as at the time there was an error in the way EE's were being used, so this brings it back to standard

# How should this be tested?

Tests

### DIFF
--- a/changelogs/fragments/async.yml
+++ b/changelogs/fragments/async.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - Add option to change async directory, and set the default to null. /tmp/.ansible_async was a workaround while the default was broken previously.
+...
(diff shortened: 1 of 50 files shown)


sean-m-sullivan:
It worked here
https://github.com/sean-m-sullivan/automation_hub_collection/actions/runs/5641297349
Not sure why it failed,

djdanielsson (approved):
LGTM

sean-m-sullivan:
@benformosa these are the tweaks, going to the Ansible default now seems prudent.

benformosa:
Thanks mate. I have a similar change ready for redhat_cop/controller_configuration. Should have a PR tomorrow.

PR #278, opened 2023-07-26
Ee repository fix

sean-m-sullivan:
# What does this PR do?
When running the ee_repository multiple times it brought up an error "Cannot make a PUT request to this endpoint" in the Create EE Repository | Wait for finish the ee_repository creation task. This fixes that by setting the correct endpoint for the object.

Also reverted the commented line on the ansible.cfg file, this should never have changed and relates more to tests.

# How should this be tested?

Automated tests


djdanielsson (approved):
LGTM

PR #284, opened 2023-08-02
Make galaxy.yml persistent and move out ansible.cfg

### LINKED ISSUE #280: ansible.cfg presence in root might break installation from repository
Collections do not ship with an `ansible.cfg` but this repository has one in the root directory. Therefore, if someone builds/installs the collection from the repository using commands like:
```bash
git clone https://github.com/ansible/automation_hub_collection
cd automation_hub_collection
ansible-galaxy collection build -vvv
ansible-galaxy install ansible-automation_hub*.tar.gz
```
then the `ansible.cfg` in the repository will override other settings.

It looks like the `ansible.cfg` is used for some tests, so it might be better to keep it elsewhere in the repository, not in the root directory.

Tompage1994:
# What does this PR do?
Makes galaxy.yml persistent and moves the template out of root. Also moves ansible.cfg from root.

This makes it installable from source

# How should this be tested?
1. Test we can install from source
2. We will need to perform a release and ensure that the galaxy.yml remains in the repo with `-devel` appended to the version that's just been released. We need to also be sure that the galaxy.yml of the released version is correct

resolves #280 

Required https://github.com/redhat-cop/ansible_collections_tooling/pull/33 to be merged


Tompage1994:
Tests will all be failing until https://github.com/redhat-cop/ansible_collections_tooling/pull/33 is merged

djdanielsson (approved):
LGTM

PR #286, opened 2023-08-09
Update links and pipeline

sean-m-sullivan:
# What does this PR do?

Fixes pipeline errors and links from the name change. The master test will still fail just as it did prior to the change.

# How should this be tested?

resolves #[number]


Tompage1994 (approved):
LGTM

.github/workflows/ci_standalone_versioned.yml
```diff
@@ -66,7 +66,7 @@ jobs:
         run: sed -i.bak 's/WITH_DEV_INSTALL=1/WITH_DEV_INSTALL=0/' galaxy_ng/.compose.env
 
       - name: set the hub ui path in .compose.env
-        run: echo "ANSIBLE_HUB_UI_PATH='/home/runner/work/automation_hub_collection/automation_hub_collection/ansible-hub-ui'" >> galaxy_ng/.compose.env
+        run: echo "ANSIBLE_HUB_UI_PATH='/home/runner/work/${{ github.event.repository.name }}/${{ github.event.repository.name }}/ansible-hub-ui'" >> galaxy_ng/.compose.env
```
Tompage1994, line 69: Pipelines failing because this change hasn't yet been reflected in the pipeline. Assume it'll work before so failing pipeline ≠ broken in this case

.github/workflows/ci_standalone_versioned.yml
```diff
@@ -92,7 +92,7 @@ jobs:
         run: ./compose up -d
 
       - name: give stack some time to spin up
-        run: sleep 240
+        run: sleep 360
```
Tompage1994: Do we often see the stack not coming up any more? Do we think this is really necessary as this will cause all pipelines to be 2 minutes longer. Could we instead replace this with something that waits for the stack to come up by polling?

.github/workflows/ci_standalone_versioned.yml
sean-m-sullivan: This change should make it so anyone with OLD repo names will work with the new one if tthey haven't changed over, however the build stack failure is fixed by this
https://github.com/ansible/galaxy_collection/actions/runs/5808103008/job/15744211764

.github/workflows/ci_standalone_versioned.yml
sean-m-sullivan: I am not sure when this changed, so reverting.

Tompage1994 (approved):

PR #291, opened 2023-08-13
add collection remote repositories

sean-m-sullivan:
# What does this PR do?

add collection remote repositories
Provides a framework on how we would rework roles/plugins to be name neutral

# How should this be tested?

Automated tests

### DIFF
--- a/changelogs/fragments/collection_remote.yml
+++ b/changelogs/fragments/collection_remote.yml
@@ -1,0 +1,6 @@
+---
+major_changes:
+  - Added collection_remote role and module to manage collection remotes. This will only work in Galaxy 4.7 or AAP 2.4.
+deprecated_features:
+  - ah_repository, ah_repository_sync modules and repository, repository_sync roles are being depreciated, they will be removed when support for AAP 2.3 ends in May of 2024.
+...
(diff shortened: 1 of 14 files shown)


sean-m-sullivan:
This should work on merge, I think something with where its pulling the versioned file from is screwing with master not skipping the one test.

as seen here
https://github.com/sean-m-sullivan/galaxy_collection/actions/runs/5850852671/job/15860832642

Tompage1994:
> This should work on merge, I think something with where its pulling the versioned file from is screwing with master not skipping the one test.
> 
> as seen here https://github.com/sean-m-sullivan/galaxy_collection/actions/runs/5850852671/job/15860832642

Yes, anything under `.github` will be whats in main to prevent horrible things being done by people raising PRs

changelogs/fragments/collection_remote.yml
```diff
@@ -0,0 +1,6 @@
+---
+major_changes:
+  - Added collection_remote role and module to manage collection remotes. This will only work in Galaxy 4.7 or AAP 2.4.
+deprecated_features:
+  - ah_repository, ah_repository_sync modules and repository, repository_sync roles are being depreciated, they will be removed when support for AAP 2.3 ends in May of 2024.
```
Tompage1994, line 5: We need to add a deprecation notice to the relevant modules and roles

PR #279, opened 2023-07-27
[WIP] Add `collection_repository` module for repository management feature

jerabekjiri:
# What does this PR do?
Enables repo management features with the `collection_repository` module.

# How should this be tested?
automated testing
 

# Is there a relevant Issue open for this?
AAH issue: [AAH-2422](https://issues.redhat.com/browse/AAH-2422)

# Other Relevant info, PRs, etc

PR #294, opened 2023-08-15
add collection repository modules and roles

sean-m-sullivan:
# What does this PR do?

add collection repository modules and roles, tweeked remote pulp naming

# How should this be tested?

Added tests

Completes and borrows from  work in #279


Tompage1994 (approved):
Looks good to me

Issue #293, opened 2023-08-14
We need to add a deprecation notice to the relevant modules and roles


Tompage1994:
We need to add a deprecation notice to the relevant modules and roles

ah_repository, ah_repository_sync modules and repository, repository_sync roles are being depreciated, they will be removed when support for AAP 2.3 ends in May of 2024.

sean-m-sullivan:
we did this,

PR #297, opened 2023-08-17
Add depreciation warnings

sean-m-sullivan:
# What does this PR do?
Address #293 and add depreciation warnings for 4 modules
This will show up when the module is run, either directly or through the roles, warnings mention both.

# How should this be tested?
The floor is lava

#293

### DIFF
--- a/changelogs/fragments/depreciation.yml
+++ b/changelogs/fragments/depreciation.yml
@@ -1,0 +1,7 @@
+---
+deprecated_features:
+  - Added warnings to ah_ee_namespace module to depreciate it and the ee_namespace role that they will be removed when support for AAP 2.2 ends in November of 2023.
+  - Added warnings to ah_group_perm module to depreciate it and the group role that uses it they will be removed when support for AAP 2.2 ends in November of 2023.
+  - Added warnings to ah_repository module to depreciate it and the repository role that they will be removed when support for AAP 2.3 ends in May of 2024.
+  - Added warnings to ah_repository_sync module to depreciate it and the repository_sync role that they will be removed when support for AAP 2.3 ends in May of 2024.
+...
--- a/plugins/modules/ah_ee_namespace.py
+++ b/plugins/modules/ah_ee_namespace.py
@@ -13,6 +13,11 @@ from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['deprecated'],
+                    'supported_by': 'community'}
+
+
 DOCUMENTATION = r"""
 ---
 module: ah_ee_namespace
@@ -24,6 +29,11 @@ description:
   - Please use the ee_repository to achieve the same functionality.
 version_added: '0.4.3'
 author: Herve Quatremain (@herve4m)
+deprecated:
+  why: The endpoint has been removed and is not supported in AAP 2.3 onwards
+  alternative: None
+  removed_in: 3.0.0
+  removed_from_collection: ansible.automation_hub
 options:
   name:
     description:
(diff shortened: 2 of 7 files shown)


changelogs/fragments/depreciation.yml
```diff
@@ -0,0 +1,7 @@
+---
+minor_changes:
```
Tompage1994: There is a heading in the changelog for this:

```suggestion
deprecated_features:
```

plugins/modules/ah_ee_namespace.py
Tompage1994, line 1: You also need to add to the documentation of the module that it is being deprecated. [This page](https://docs.ansible.com/ansible/latest/dev_guide/module_lifecycle.html#module-lifecycle) explains it but you need to add:

```
deprecated:
  removed_in: <version target>
  removed_by_date: <date>  # Mutually exclusive with removed_in 
  why: <optional>
  alternative: <optional>  # Something like: Use M(x.y.z) instead.
```

My suggestion is we go for something here like:

```
deprecated:
  removed_by_date: November 2023
  why: The endpoint has been removed and is not supported in AAP 2.3 onwards
  alternative: Use M(<alt module>) instead.
```

Alternatively we can go for `removed_in: 3.0.0` as we will be releasing 2.0.0 soon and we will need to major bump the version when removing a module.

plugins/modules/ah_group_perm.py
```diff
@@ -165,6 +165,9 @@ def main():
 
     # Create a module for ourselves
     module = AHAPIModule(argument_spec=argument_spec, supports_check_mode=True)
+    module.warn("This module 'ah_group_perm' and the group permision part of the role 'group' will be removed when support for AAP 2.2 "
+                "ends in November of 2023. It was repplaced with role permisions in 2.3."
```
Tompage1994:
```suggestion
                "ends in November of 2023. It was replaced with role permissions in 2.3."
```

plugins/modules/ah_ee_namespace.py
```diff
@@ -208,6 +208,9 @@ def main():
 
     # Create a module for ourselves
     module = AHAPIModule(argument_spec=argument_spec, supports_check_mode=True)
+    module.warn("This role 'ee_namespace' and module 'ah_ee_namespace' will be removed when support for AAP 2.2 ends in November of 2023. "
+                "It was removed in 2.3."
```
Tompage1994: Haven't checked but is there some endpoint to replace this? Is this a new module we need to write?

plugins/modules/ah_ee_namespace.py
Tompage1994: Obviously needed for each module

Tompage1994 (changes requested):
A few changes and suggestions

plugins/modules/ah_ee_namespace.py
sean-m-sullivan: No, it was just completely removed, outright, no replacement

plugins/modules/ah_ee_namespace.py
```diff
@@ -24,6 +29,10 @@
   - Please use the ee_repository to achieve the same functionality.
 version_added: '0.4.3'
 author: Herve Quatremain (@herve4m)
+deprecated:
+  removed_by_date: November 2023
+  why: The endpoint has been removed and is not supported in AAP 2.3 onwards
+  alternative: Use M(<alt module>) instead.
```
Tompage1994: Is it better to just say there is no alternative at present?

Tompage1994 (approved):
Just one comment and up to you whether to change it

plugins/modules/ah_ee_namespace.py
sean-m-sullivan: I'm working on getting these working

sean-m-sullivan:
Okay I think I have it now. This passes sanity,

djdanielsson (approved):
LGTM

Issue #179, opened 2022-12-07
Add new module for adding group roles to replace the group_perms module


Tompage1994:
# Summary

Add new module for adding group roles to replace the group_perms module

The endpoint to do this is something like: /api/pulp/api/v3/groups/{id}/add_role/

# Issue Type

- Feature request

djdanielsson:
ok I added discussions so now feature requests can be put there if we want instead of issues, Idk which I like better.

Tompage1994:
> ok I added discussions so now feature requests can be put there if we want instead of issues, Idk which I like better.

Lets go with discussions for now. My only issue is whether they will be visible enough there. We can change if required

sean-m-sullivan:
This is going to be on hold, as it is dependent on https://issues.redhat.com/browse/AAH-2080

Until that endpoint works, this will not and is impossible to work on.

marszip:
Are there any updates to this issue, the Red Hat issue which was blocking seems to be resolved. Not sure if that is the case, but we would like to manage our roles with this collection.

sean-m-sullivan:
its still and issue, but its in our backlog

EamonTracey:
+1 on this issue. I would like to be able to test ah_group_perm (or a new role module) in the galaxy_ng CI workflows

sean-m-sullivan:
Closing as this was fixed in #299

PR #299, opened 2023-08-20
Add group role management and other improvements

sean-m-sullivan:
# What does this PR do?

major_changes:
  - Added group_roles role and module to manage and attach roles to groups. Glbal role assignment will only work in Galaxy 4.6 or AAP 2.3 or later, while the targeted role assignment will work in Galaxy 4.7 or AAP 2.4 and above.

minor_changes:
  - Updated output in multiple modules to provide more information.
  - Updated server version lookup to remove the dev portion if working on a developmental version of Galaxy.
  - Updated ah_groups module to use new Pulp endpoint. The change should be seamless.
  - added more permissions to the ah_role modules. Consult your web UI to see what roles are available specific to your version.
  - added more roles to the dispatch role

# How should this be tested?

Automated tests have been added.

#179

### DIFF
--- a/changelogs/fragments/group_roles.yml
+++ b/changelogs/fragments/group_roles.yml
@@ -1,6 +1,6 @@
 ---
 major_changes:
-  - Added group_roles role and module to manage and attach roles to groups. Glbal role assignment will only work in Galaxy 4.6 or AAP 2.3 or later, while the targeted role assignment will work in Galaxy 4.7 or AAP 2.4 and above.
+  - Added group_roles role and module to manage and attach roles to groups. Global role assignment will only work in Galaxy 4.6 or AAP 2.3 or later, while the targeted role assignment will work in Galaxy 4.7 or AAP 2.4 and above.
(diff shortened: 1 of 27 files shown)


changelogs/fragments/group_roles.yml
```diff
@@ -0,0 +1,10 @@
+---
+major_changes:
+  - Added group_roles role and module to manage and attach roles to groups. Glbal role assignment will only work in Galaxy 4.6 or AAP 2.3 or later, while the targeted role assignment will work in Galaxy 4.7 or AAP 2.4 and above.
```
Tompage1994:
```suggestion
  - Added group_roles role and module to manage and attach roles to groups. Global role assignment will only work in Galaxy 4.6 or AAP 2.3 or later, while the targeted role assignment will work in Galaxy 4.7 or AAP 2.4 and above.
```

Tompage1994 (approved):
LGTM

PR #306, opened 2023-08-24
Convert to Galaxy namespace

sean-m-sullivan:
# What does this PR do?
Convert to Galaxy namespace
Update to work with earlier versions of python, removing some 3.8+ only code.

# How should this be tested?
Automated tests have been updated, as have sanity.


djdanielsson (approved):
LGTM

PR #318, opened 2023-10-03
fix: correct variable to create namespace

jon4hz:
Currently the dispatch role tries to create collection namespaces based on the `ah_ee_namespaces` variable. The correct variable would be `ah_namespaces`.


sean-m-sullivan (approved):
good catch, Lgtm

PR #335, opened 2023-11-16
Fix version comparison

### LINKED ISSUE #334: Version Comparison is broken when galaxy version >= 4.10.0
Error: https://github.com/ansible/galaxy_ng/actions/runs/6892485007/job/18749880822?pr=1981

"msg": "This module requires private automation hub version 4.6 or later. Your version is 4.10.0"

This happens because we are bumping galaxy_ng version to "4.10.0" and the comparison on modules are using string comparison.

https://github.com/search?q=repo%3Aansible%2Fgalaxy_collection%20%22this%20module%20requires%22&type=code

## Proposal

Use one of:

Needs packaging as a requirement (is that already in Ansible)?
```python
>>> # pip install packaging
>>> from packaging import version
>>> version.parse("2.3.1") < version.parse("10.1.2")
True
>>> version.parse("1.3.a4") < version.parse("10.1.2")
True
>>> isinstance(version.parse("1.3.a4"), version.Version)
True
>>> isinstance(version.parse("1.3.xy123"), version.LegacyVersion)
True
```

OR

```python
>>> from pkg_resources import parse_version
>>> parse_version('1.9.a.dev') == parse_version('1.9a0dev')
True
>>> parse_version('2.1-rc2') < parse_version('2.1')
True
>>> parse_version('0.6a9dev-r41475') < parse_version('0.6a9')
True
```

rochacbruno:
Compara versions using ansible.module_utils.compat.version.LooseVersion

closes #334

### DIFF
--- a/plugins/modules/ah_ee_image.py
+++ b/plugins/modules/ah_ee_image.py
@@ -9,6 +9,7 @@
 
 
 from __future__ import absolute_import, division, print_function
+from packaging.version import Version
 
 __metaclass__ = type
 
@@ -121,7 +122,7 @@ def main():
 
     # Only recent versions support execution environment
     vers = module.get_server_version()
-    if vers < "4.3.2":
+    if Version(vers) < Version("4.3.2"):
         module.fail_json(msg="This module requires private automation hub version 4.3.2 or later. Your version is {vers}".format(vers=vers))
 
     # Process the object from the Pulp API (delete or create)
--- a/plugins/modules/ah_ee_image.py
+++ b/plugins/modules/ah_ee_image.py
@@ -9,7 +9,6 @@
 
 
 from __future__ import absolute_import, division, print_function
-from ansible.module_utils.compat.version import LooseVersion as Version
 
 __metaclass__ = type
 
@@ -91,6 +90,7 @@ EXAMPLES = r"""
 
 RETURN = r""" # """
 
+from ansible.module_utils.compat.version import LooseVersion as Version
 from ..module_utils.ah_api_module import AHAPIModule
 from ..module_utils.ah_ui_object import AHUIEEImage
 from ..module_utils.ah_pulp_object import AHPulpEERepository
--- a/plugins/modules/ah_ee_image.py
+++ b/plugins/modules/ah_ee_image.py
@@ -9,7 +9,7 @@
 
 
 from __future__ import absolute_import, division, print_function
-from packaging.version import Version
+from ansible.module_utils.compat.version import LooseVersion as Version
 
 __metaclass__ = type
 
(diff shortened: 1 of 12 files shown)


rochacbruno:
packaging is added by ansible https://github.com/ansible/ansible/blob/devel/requirements.txt#L9

Not sure why the CI is failing, do we need to include it elsewhere?

sean-m-sullivan (approved):
Clean, thanks!

rochacbruno:
@sean-m-sullivan CI is failing because the packaging dependency cannot be found

sean-m-sullivan:
it looks like this dependency is not part of the ansible install. need it in requirements.txt in the base directory.

rochacbruno:
ALternativelly we can create our own implementation of the class `Version` and not depend on external library.

```python
class Version:
    def __init__(self, version_str):
        components = list(map(int, version_str.split('.')))
        self.major = components[0]
        self.minor = components[1] if len(components) > 1 else 0
        self.patch = components[2] if len(components) > 2 else 0

    def __str__(self):
        return f"{self.major}.{self.minor}.{self.patch}"

    def __eq__(self, other):
        if not isinstance(other, Version):
            return NotImplemented
        return (self.major, self.minor, self.patch) == (other.major, other.minor, other.patch)

    def __lt__(self, other):
        if not isinstance(other, Version):
            return NotImplemented
        return (self.major, self.minor, self.patch) < (other.major, other.minor, other.patch)

    def __le__(self, other):
        if not isinstance(other, Version):
            return NotImplemented
        return (self < other) or (self == other)

    def __gt__(self, other):
        if not isinstance(other, Version):
            return NotImplemented
        return not (self <= other)

    def __ge__(self, other):
        if not isinstance(other, Version):
            return NotImplemented
        return not (self < other)

    def __ne__(self, other):
        if not isinstance(other, Version):
            return NotImplemented
        return not (self == other)

# Example usage:
version1 = Version("4.2")
version2 = Version("5.1.3")

print(version1 < version2)  # Output: True
print(version1 >= version2)  # Output: False

```

plugins/modules/ah_ee_image.py
```diff
@@ -9,6 +9,7 @@
 
 
 from __future__ import absolute_import, division, print_function
+from ansible.module_utils.compat.version import LooseVersion as Version
```
Tompage1994: sanity check is stating that this needs to be with the imports further down below the documentation segment

Tompage1994: Plus the same for all the other modules

Tompage1994 (changes requested):

rochacbruno:
Trying a simpler implementation on https://github.com/ansible/galaxy_collection/pull/337

rochacbruno:
Closing in favor of #337

PR #337, opened 2023-11-17
Fix version comparison (impl 2)

### LINKED ISSUE #334: Version Comparison is broken when galaxy version >= 4.10.0

rochacbruno:
fix #334

alternative implementarion for #335

### DIFF
--- a/plugins/module_utils/ah_api_module.py
+++ b/plugins/module_utils/ah_api_module.py
@@ -17,6 +17,7 @@ import json
 import time
 
 from ansible.module_utils.basic import AnsibleModule, env_fallback
+from ansible.module_utils.compat.version import LooseVersion as Version
 from ansible.module_utils._text import to_bytes, to_text
 
 from ansible.module_utils.six.moves.urllib.parse import urlparse, urlencode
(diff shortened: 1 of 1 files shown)


Tompage1994 (commented):
Will this not still be broken in all the module files?

sean-m-sullivan (approved):

rochacbruno:
@Tompage1994 no because it is allowed to do `Version("1.2.3") < "1.2.3"` the right hand can be a string.

PR #342, opened 2023-11-27
user: Fix password type field in README

dsavineau:
The `password` type was set to bool instead of str.


djdanielsson (approved):
good catch

PR #346, opened 2023-12-20
collection_remote: set sync_dependencies default to PAH default

dbk-rabel:
# What does this PR do?

The default of sync_dependencies in PAH is True. So the ah_configuration collection should also use True as the default value.

### DIFF
--- a/plugins/modules/collection_remote.py
+++ b/plugins/modules/collection_remote.py
@@ -127,7 +127,7 @@ options:
     description:
       - Whether to download depenencies when syncing collections.
     type: bool
-    default: False
+    default: True
   proxy_url:
     description:
       - Proxy URL to use for the connection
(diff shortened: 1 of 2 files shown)


djdanielsson (changes requested):
I'm confused, in your PR comment you said change to False but the change is to True

dbk-rabel:
> I'm confused, in your PR comment you said change to False but the change is to True

Oh yeah. Thanks for the hint. I fixed it now.

dbk-rabel:
To clarify: The change is to True. The commit was correct. I edited the PR comment.

djdanielsson (approved):
LGTM

djdanielsson:
I am going to merge this because the CI failing isn't related to this change from what I can tell

PR #359, opened 2024-02-07
Fix documentation for group_roles role

Tompage1994:
# What does this PR do?
Fixes the documentation for the group_roles role. Previously mentioned ah_groups which is the wrong variable name for the role. 

# How should this be tested?
N/A doc change

resolves internal Jira AAP-20247

related #357


djdanielsson (approved):
LGTM

PR #363, opened 2024-02-20
Fixed issue in all roles where AH_HOST was not being defaulted to if no variable set

### LINKED ISSUE #362: AH_HOST not being fallen back to in all roles
# Summary
In all roles the AH_HOST env var is not being fallen back to when no variable is set.

# Issue Type

- Bug Report

# Ansible, Collection, Private Automation Hub details
latest of all

# OS / ENVIRONMENT
MacOS

# Desired Behavior
Expect no variable to be set but AH_HOST env var to be used.

# Actual Behavior

```
TASK [infra.ah_configuration.namespace : Add Automation Hub Namespace] **************************************************************************************************************************************************************************************************************************************************************************
fatal: [automation_hub]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'ah_hostname' is undefined. 'ah_hostname' is undefined
```

# STEPS TO REPRODUCE
Set AH_HOST env var and run any role but don't set `ah_host` or `ah_hostname` and see it fails

Tompage1994:
# What does this PR do?
Default omits the hostname option on the role so we fallback to the env var

# How should this be tested?
Env var can be used.

We don't have any ENV var testing in the CI. Maybe we should...

resolves #362 


djdanielsson (approved):
LGTM

Issue #358, opened 2024-02-07
CI Failing to build AH stack


Tompage1994:
# Summary
CI currently failing on 'build stack' step

# Issue Type

- Bug Report

# Ansible, Collection, Private Automation Hub details
All versions

# OS / ENVIRONMENT
CI - GH Actions

# Desired Behavior
Stack builds

# Actual Behavior

```
Run make docker/build
INFO: Using compose profile standalone
INFO: galaxy_ng packages installed from source
INFO: Image suffix is unset
INFO: Volume suffix is unset
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.[8](https://github.com/ansible/galaxy_collection/actions/runs/7532803957/job/20504121587#step:16:9).18/x64/bin/docker-compose", line 8, in <module>
    sys.exit(main())
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/compose/cli/main.py", line 81, in main
    command_func()
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/compose/cli/main.py", line 200, in perform_command
    project = project_from_options('.', options)
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/compose/cli/command.py", line 60, in project_from_options
    return get_project(
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/compose/cli/command.py", line 152, in get_project
    client = get_client(
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/compose/cli/docker_client.py", line 41, in get_client
    client = docker_client(
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/compose/cli/docker_client.py", line [12](https://github.com/ansible/galaxy_collection/actions/runs/7532803957/job/20504121587#step:16:13)4, in docker_client
    kwargs = kwargs_from_env(environment=environment, ssl_version=tls_version)
TypeError: kwargs_from_env() got an unexpected keyword argument 'ssl_version'
make: *** [Makefile:83: docker/build] Error 1
Error: Process completed with exit code 2.
```

PR #364, opened 2024-02-20
Fix documentation for validate_certs

Tompage1994:
# What does this PR do?
Fixes documentation which says that the default for validate certs is `false` and the arg_spec says `true` when in fact it is omitted.

# How should this be tested?
N/A - Doc change

N/A

### DIFF
--- a/roles/collection/README.md
+++ b/roles/collection/README.md
@@ -11,7 +11,7 @@ An Ansible Role to update, or destroy Automation Hub Collections.
 |`ah_host`|""|yes|URL to the Automation Hub or Galaxy Server. (alias: `ah_hostname`)|127.0.0.1|
 |`ah_username`|""|yes|Admin User on the Automation Hub or Galaxy Server.||
 |`ah_password`|""|yes|Automation Hub Admin User's password on the Automation Hub Server.  This should be stored in an Ansible Vault at vars/tower-secrets.yml or elsewhere and called from a parent playbook.||
-|`ah_validate_certs`|`False`|no|Whether or not to validate the Ansible Automation Hub Server's SSL certificate.||
+|`ah_validate_certs`|`true`|no|Whether or not to validate the Ansible Automation Hub Server's SSL certificate.||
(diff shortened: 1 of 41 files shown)


roles/collection/README.md
```diff
@@ -11,7 +11,7 @@ An Ansible Role to update, or destroy Automation Hub Collections.
 |`ah_host`|""|yes|URL to the Automation Hub or Galaxy Server. (alias: `ah_hostname`)|127.0.0.1|
 |`ah_username`|""|yes|Admin User on the Automation Hub or Galaxy Server.||
 |`ah_password`|""|yes|Automation Hub Admin User's password on the Automation Hub Server.  This should be stored in an Ansible Vault at vars/tower-secrets.yml or elsewhere and called from a parent playbook.||
-|`ah_validate_certs`|`False`|no|Whether or not to validate the Ansible Automation Hub Server's SSL certificate.||
+|`ah_validate_certs`|""|no|Whether or not to validate the Ansible Automation Hub Server's SSL certificate.||
```
djdanielsson: Wouldn't it be default true then?

djdanielsson (changes requested):
I think it should state that it is default true because that is how it will work at the module level

djdanielsson:
Also can you fix the dead link so the linters do not error

roles/collection/README.md
Tompage1994: Yes makes sense. I was going for what the role defaults but actually by convention we say what the module defaults here. 

I will keep the arg_spec with no default because technically the default is not on that variable.

Tompage1994:
> Also can you fix the dead link so the linters do not error

weirdly this link does work but seems to provide `403` in CI. I wonder whether SO have started doing something to prevent CI from getting content from them. I've just made it into a comment rather than a link so should be better now

Tompage1994:
We have #358 open for the failing pipelines so we can go ahead and ignore the CI for this type of change for now.

Issue #365, opened 2024-02-22
Collection won't install from git repo


kbreit-insight:
# Summary

I have a requirements.yml file which installs this collection from the repo and it's failing. I am unsure if it's a problem with the collection or ansible-galaxy/requirements.yml.

# Issue Type

- Bug Report

# Ansible, Collection, Private Automation Hub details

```console (paste below)
+ ansible-galaxy --version
ansible-galaxy [core 2.15.6]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/runner/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  ansible collection location = /home/runner/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible-galaxy
  python version = 3.9.18 (main, Sep  7 2023, 00:00:00) [GCC 11.4.1 20230605 (Red Hat 11.4.1-2)] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True

```

- ansible installation method: one of source, pip, OS package, EE

# OS / ENVIRONMENT

Ubuntu

# Desired Behavior

It should build with the collection.

# Actual Behavior
```
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/ansible/galaxy/collection/concrete_artifact_manager.py", line 301, in get_direct_collection_meta
    return self._artifact_meta_cache[collection.src]
KeyError: None
 
During handling of the above exception, another exception occurred:
 
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/ansible/cli/__init__.py", line 659, in cli_executor
    exit_code = cli.run()
  File "/usr/lib/python3.9/site-packages/ansible/cli/galaxy.py", line 719, in run
    return context.CLIARGS['func']()
  File "/usr/lib/python3.9/site-packages/ansible/cli/galaxy.py", line 119, in method_wrapper
    return wrapped_method(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/ansible/cli/galaxy.py", line 1338, in execute_install
    requirements = self._parse_requirements_file(
  File "/usr/lib/python3.9/site-packages/ansible/cli/galaxy.py", line 818, in _parse_requirements_file
    requirements['collections'] = [
  File "/usr/lib/python3.9/site-packages/ansible/cli/galaxy.py", line 819, in <listcomp>
[...]

PR #367, opened 2024-02-29
Remove bindep

sean-m-sullivan:
# What does this PR do?

Removing Bindep.txt and its python requirement. This should be a given with ansible.

# How should this be tested?

#365


changelogs/fragments/remove_bindep.yml
```diff
@@ -0,0 +1,4 @@
+---
+major_changes:
+  - removed bindep.txt as the only requirement was python. Ansible requires python, it shouldn't be needed in our bindep.
```
Tompage1994, line 3: Is this really a major change? I think at most its a minor, and may even be a bug fix? In reality this isn't going to break anything or fundamentally add anything so should stay lower

Tompage1994 (changes requested):

changelogs/fragments/remove_bindep.yml
sean-m-sullivan: updated to bugfix

djdanielsson (approved):
LGTM

Tompage1994 (approved):

PR #370, opened 2024-03-21
updating license and ansible min to match everywhere

djdanielsson:
updating license and ansible min to match everywhere


sean-m-sullivan (approved):
LGTM

Issue #311, opened 2023-09-12
Weird issue with user module


nodje:
Using 2.0.2:

This config fails:
```
---
ah_users:
  - username: service
    is_superuser: true
    password: !vault |
      $ANSIBLE_VAULT;1.1;AES256
      30643930336231306161343931396539396530646631623966646638643963333832353065333632
      3166386438303265336566336330663933313366363530310a643335653364643462313166333539
      30313266363834633037366639623365356634613562613038396661303961336332386435643832
      3231623761363332610a663431313531656235373837636138366135313130653237346537626533
      6166
  - username: DBA
    groups:
      - DBA
    append: false
    password: !vault |
      $ANSIBLE_VAULT;1.1;AES256
      32626137336230373934633263623830666334376434336637363138353732323537313366643231
      3232353365633264373036326465303036373866326162310a353336363432313137343937643765
      30613463666133663830643562333430623363393438376433366430386331373464363764613339
      3566653562333835380a646433393333623865653165646362333663636138303663653233636236
      3533
...
```
with log:
```
TASK [infra.ah_configuration.user : Create User | Wait for finish the user creation] *********************************************************************************************************************************************************************
FAILED - RETRYING: [localhost]: Create User | Wait for finish the user creation (50 retries left).
changed: [localhost] => (item={'failed': 0, 'started': 1, 'finished': 0, 'ansible_job_id': '985190343965.294533', 'results_file': '/root/.ansible_async/985190343965.294533', 'changed': False, '__user': {'username': 'service', 'is_superuser': True, 'password': 'stagingbis'}, 'ansible_loop_var': '__user'}) => {"__users_job_async_result_item": {"__user": {"is_superuser": true, "password": "stagingbis", "username": "service"}, "ansible_job_id": "985190343965.294533", "ansible_loop_var": "__user", "changed": false, "failed": 0, "finished": 0, "results_file": "/root/.ansible_async/985190343965.294533", "started": 1}, "ansible_job_id": "985190343965.294533", "ansible_loop_var": "__users_job_async_result_item", "attempts": 2, "changed": true, "finished": 1, "id": 3, "name": "service", "results_file": "/root/.ansible_async/985190343965.294533", "started": 1, "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": [], "type": "user", "warnings": ["The field password of user service has encrypted data and may inaccurately report task is changed."]}
[...]

PR #383, opened 2024-04-17
Handle errors correctly

Tompage1994:
# What does this PR do?
Fixes an issue where errors aren't correctly handled

# How should this be tested?
CI should work better and return a real error message if the user module fails (probably telling us the password is too short)

No

related #311 


djdanielsson (approved):
LGTM

PR #395, opened 2024-05-18
Replace order of roles

laurobmb:
during the installation of the ansible automation platform an error occurs when the installation uses postinstall, in the variables in the defaults/main.yml folder the playbook tries to configure the ah_ee_repositories before the ah_ee_registries this causes the installation to break, causing the roles and registries to be inverted are configured first and the repositories follow.

the test was carried out by installing ansible automation platform 2.4 containerized on a rhel 9

This issue corrects a broken installation problem if you use AH's postinstall

resolves #[number]


djdanielsson (changes requested):
it failed the pre-commit checks, please fix these

Tompage1994 (approved):

PR #399, opened 2024-05-23
Add ah_configuration_{..}async_timeout parameters

### LINKED ISSUE #398: Timeout parameter in repository_sync does not work; Was it meant to be an async parameter?
Hi.

https://github.com/ansible/galaxy_collection/blob/devel/roles/repository_sync/README.md?plain=1#L60 says, that the timeout parameter for repository_sync does the following: "If waiting for the project to update this will abort after this amount of seconds."

It doesn't. Actually it doesn't seem to have a lot of effect.

What would be useful instead (and maybe was intended here), would be to be able to increase the async parameter: https://github.com/ansible/galaxy_collection/blob/devel/roles/repository_sync/tasks/main.yml#L32C10-L32C14

Maybe it would also make sense to have a globale parameter called ah_configuration_async_timeout or something to set this.

But especially for repository_sync it is necessary, because you can easily get above 1000 seconds.

Yours
  David

dbk-rabel:
Fixes #398

Add ah_configuration_{..}async_timeout parameters to make async parameter (timeout for async tasks) configurable.

### DIFF
--- a/roles/collection/README.md
+++ b/roles/collection/README.md
@@ -34,6 +34,22 @@ These are the sub options for the vars `ah_collections` which are dictionaries w
 The `ah_configuration_async_dir` variable sets the directory to write the results file for async tasks.
 The default value is set to  `null` which uses the Ansible Default of `/root/.ansible_async/`.
 
+### Asynchronous Retry Variables
+
+The following Variables set asynchronous retries for the role.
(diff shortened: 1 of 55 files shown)


djdanielsson (approved):
a quick scan looks good to me

Tompage1994 (approved):

PR #409, opened 2024-06-18
Test CI

bmclaughlin:
# What does this PR do?
Attempting to fix CI errors for master on the standalone and dab_jwt workflows.  Other failing stable-xxx workflows to be addressed in future PRs.

# How should this be tested?

No

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -11,15 +11,15 @@ assignees: ''
 <!--- Verify first that your issue is not already reported on GitHub -->
 <!--- Also test if the latest release are affected -->
 
-# Summary
+## Summary
 
 <!--- Explain the problem briefly below -->
 
-# Issue Type
+## Issue Type
 
 - Bug Report
 
-# Ansible, Collection, Private Automation Hub details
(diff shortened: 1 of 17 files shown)


.github/workflows/ci_standalone_versioned.yml
```diff
@@ -58,6 +58,9 @@ jobs:
           echo "OCI_VERBOSE=1" >> $GITHUB_ENV
           echo "GH_DUMP_LOGS=1" >> $GITHUB_ENV
           echo "GH_TEARDOWN=0" >> $GITHUB_ENV
+          echo "AH_HOST=http://localhost:55001/" >> $GITHUB_ENV
+          echo "AH_USERNAME=admin" >> $GITHUB_ENV
+          echo "AH_PASSWORD=admin" >> $GITHUB_ENV
```
bmclaughlin: @sean-m-sullivan Is this the right place to set the environment variables?  This change doesn't appear to be recognized (no idea if it's cached etc, very weird) by github and I think it may be the root cause the `standalone` CI is failing. 

Testing these playbooks locally with set env vars the playbooks only show failures due to my not passing in all the variables (ie: `redhat_catalog_username` etc).

sean-m-sullivan (approved):
Looks good to me

PR #419, opened 2024-07-19
Add empty_requirements to collection_remote

### LINKED ISSUE #413: Unable to remove requirements.yml from a remote.
### Discussed in https://github.com/ansible/galaxy_collection/discussions/412

<div type='discussions-op-text'>

<sup>Originally posted by @smatterchew June 26, 2024</sup>
Hello.

I am trying to remove a `requirements.yml` from a remote in Private Automation Hub v4.9.1 using `infra.ah_configuration.collection_remote`. I've tried setting the `requirements: []` to an empty list and the task succeeded, but the remote still had a collection list. Then I tried to set `requirements_file: empty_requirements_file.yml` and the empty file wasn't found in the files directory off the root of the role directory. I also tried to specify the path to the file as `requirements_file: ./files/empty_requirements_file.yml` and that also failed.

1. How come the first process of defining an empty list did not remove the requirements file? It didn't even end up as an empty list. The existing collection was still there.
2. Where in the role can I store the empty requirements file so the task module can find it? I don't want to host it on the execution node outside the Execution Environment.
3. Can we just provide a state (default: present) switch for the requirements file to have the module delete the requirements file?

Thank you in advance for your help.</div>

saito-hideki:
# What does this PR do?

Add new `empty_requirements` parameter to the `collection_remote` module clear requirements.yml contents.

# How should this be tested?

```
  tasks:
  - galaxy.galaxy.collection_remote:
      ah_host: "galaxy.example.com"
      ah_username: "<YOUR_ADMIN_USERNAME>"
      ah_password: "<YOUR_PASSWORD>"
      name: "example-remote"
      empty_requirements: true
      sync_dependencies: true
      validate_certs: false
      state: present
```

* Fixes #413

* None

### DIFF
--- a/changelogs/fragments/add_empty_requirements.yml
+++ b/changelogs/fragments/add_empty_requirements.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - add new parameter empty_requirements to collection_remote. This parameter clears the contents by setting null in requirements.yml (#413).
--- a/plugins/modules/collection_remote.py
+++ b/plugins/modules/collection_remote.py
@@ -60,6 +60,13 @@ options:
     description:
       - A yaml requirements file to download from remote.
     type: str
+  empty_requirements:
+    description:
+      - Empty the contents of requirements.yml for collection remote.
+      - If a new collection remote is created, the content of requirements.yml will be empty.
+      - If a collection remote already exists, empty the contents of requirements.yml.
+    type: bool
+    default: False
   username:
     description:
       - Username to authenticate to the remote repository.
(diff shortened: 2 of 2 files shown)


djdanielsson:
I am not sure if I like this, I think we might just need to tweak the code (role and/or python) to accept "" and thus remove any requirements that are currently in the remote. Thoughts @sean-m-sullivan @Tompage1994

Tompage1994 (commented):
My gut feeling is that this isn't a good way to do this. Ideally we should just be able to set `requirements: []` and it work. I'm intreaged as to why this doesn't work and how it does actually work through the API

Tompage1994:
> My gut feeling is that this isn't a good way to do this. Ideally we should just be able to set `requirements: []` and it work. I'm intreaged as to why this doesn't work and how it does actually work through the API

I've found a better way to do this where this bug is simply fixed. I'm going to close this PR, but thanks for the contribution @saito-hideki

PR #420, opened 2024-07-19
Fixes issue in collection_remotes module where the requirements could not be blanked out

### LINKED ISSUE #413: Unable to remove requirements.yml from a remote.

Tompage1994:
# What does this PR do?
Fixes issue where the `[]` which might be set against the requiremnets option is ignored.

# How should this be tested?
Attempt to remove requirements from a remote as follows:

```
- name: Delete requirements
      infra.ah_configuration.collection_remote:
        name: test
        url: https://galaxy.ansible.com
        requirements: []
```

resolves #413 

### DIFF
--- a/changelogs/fragments/collection_remote_requirements.yml
+++ b/changelogs/fragments/collection_remote_requirements.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - Fixes issue in collection_remotes module where the requriements could not be blanked out
+...
--- a/plugins/modules/collection_remote.py
+++ b/plugins/modules/collection_remote.py
@@ -230,9 +230,12 @@ def main():
     new_fields = {}
 
     requirements = module.params.get("requirements")
-    if requirements:
-        requirements_content = "\n  - ".join(requirements)
-        new_fields["requirements_file"] = "---\ncollections:\n  - " + requirements_content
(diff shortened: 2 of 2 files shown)


djdanielsson (approved):
LGTM

PR #424, opened 2024-08-08
Fix CI workflow issue

bmclaughlin:
# What does this PR do?
Get the workflows running again.  The input definitions were missing for ci_workflow in the local reusable github actions.  This PR adds them.

# How should this be tested?
Due to the fix being in the reusable local workflow files, the ci_standalone and ci_dab_jwt workflows still won't run on this PR.  I'm opening an additional PR in a few minutes that should have the workflows running due to this PR. 

# Is there a relevant Issue open for this?
No

# Other Relevant info, PRs, etc
All other PRs recently that the CI hasn't even started. :(

PR #425, opened 2024-08-08
Fix the tests in testing_playbook_ee_repository

bmclaughlin:
# What does this PR do?
Fixes the tests in testing_playbook_ee_repository, there were a few errors that caused failures. 

# How should this be tested?
Unfortunately I think it'll have to be verified manually, as the playbook prior to this one fails.

AAP-28358

If the standalone and dab_jwt CI workflows run (not expected to pass yet), #424 was successful.


matoval (approved):
Tested locally, testing_playbook_ee_repository all passed with no failures

djdanielsson (approved):
LGTM

PR #436, opened 2024-10-02
removing modules

djdanielsson:
# What does this PR do?
removes modules now that ansible.hub collection exists

# How should this be tested?
shouldn't be needed

# Is there a relevant Issue open for this?
no

# Other Relevant info, PRs, etc
none

PR #437, opened 2024-10-02
Remove modules

djdanielsson:
updating readme and changeling to cover the changes done in #436


.github/workflows/pre-commit.yml
```diff
@@ -18,4 +18,5 @@ jobs:
       collection_name: galaxy
       collection_version: 1.1.1-devel
       collection_repo: https://github.com/ansible/galaxy_collection
+      collection_dependencies: ansible.hub
```
Tompage1994: I believe this should fix the broken CI here so I'm happy with this

Tompage1994 (approved):

.github/workflows/pre-commit.yml
Tompage1994: Oh this wasn't even part of this diff, GH going weird. Still broken...

djdanielsson: yea idk

PR #444, opened 2025-02-05
Hide token generated by "Get token" tasks

w4hf:
# What does this PR do?
Use the existing `ah_configuration_*_secure_logging` variables to give the possibility to hide the generated token

Added `no_log: "{{ ah_configuration_X_secure_logging }}"` to every `Get Token` task
Created a new variable `ah_configuration_offline_sync_secure_logging` for the `offline_sync` role

# How should this be tested?
Set `ah_configuration_X_secure_logging` of a role which has a `Get Token` task to `true` and see that the authentication token is now hidden in the output of the task `Get Token`

N/A

The code in this repo is often used in pipelines where the job log is public.
It is a good practice to hide the token which usually have superadmin role.


sean-m-sullivan (approved):
LGTM, uses the built in function for this,

